PR #8, opened 2017-04-30
Use Ocamlbuild, menhir, and opam

objmagic:
1. Switch to use ocamlbuild, and Menhir parser generator. This largely cleans up the Makefile.
2. Use opam to allow easy local install and distribution.
3. Update README on how to use opam to locally install `macrodown`


gfngfn:
I checked it works correctly. Thank you for introducing OCamlBuild to our project in a kindly way!

PR #16, opened 2018-02-10
bump camel version to >= 4.06.0

objmagic:
Cannot install with new camel (>= `4.06.0`)

https://github.com/gfngfn/camlpdf/pull/1 should be merged before this PR gets merged.

Resolves #15

PR #25, opened 2018-02-12
Add more explicit installation instructions

nekketsuuu:
This PR add more explicit installation instructions into README.

I tested the instructions for Ubuntu on plain ubuntu:16.04 image of Docker.
However, I haven't tested those for macOS because I don't have any Mac.
So before merging, please test it if you can :pray:. Feel free to edit instructions.

Also, this instructions is for OCaml 4.05.0, not 4.06.0.
So after merging #16, this should be edited.

### DIFF
--- a/README-ja.md
+++ b/README-ja.md
@@ -9,13 +9,70 @@
 
 本ソフトウェアは2017年度IPA未踏事業の1プロジェクトとして支援のもと開発されました。
 
-## インストール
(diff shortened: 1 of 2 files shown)


gfngfn:
Confirmed that invoking commands on macOS 10.12 according to the instructions in `README`(` `|`-ja`)`.md` produces correct installation. Thank you.

PR #30, opened 2018-02-14
Add a command line option --type-check-only

sakas--:
When `--type-check-only` is set SATySFi stops after type checking.
This will come in handy when we implement some editor support features.
I also modified the log message so that it emphasizes the fact that the type checking has succeeded.

@gfngfn could you review the code?
I tried to follow your coding style, but I'm still not familiar with the codebase.


qnighy:
@sakas-- do you still have interests in it? There are conflicts with the current master.

sakas--:
If needed, I can fix and rebase it.
Or simply close this issue if no one want's this feature.

qnighy:
Considering the thumbs you received, it's reasonable to think of it as useful.

sakas--:
I've resolved the conflicts.

gfngfn:
I’m terribly sorry for the extremely late response… Confirmed it works fine. Thank you for the enhancement!

PR #32, opened 2018-02-15
add some math commands to math.satyh

ToruNiina:


KilledByNLP:
I think delta(δ) is also needed

nekketsuuu:
U+1D700 is 'MATHEMATICAL ITALIC SMALL EPSILON' 𝜀. There should be some discussion about the `\epsilon` v.s. `\varepsilon` problem (If you don't know this, see [this Q&A](https://tex.stackexchange.com/q/98013/114888)). I want to see @gfngfn's opinion.

Also, the same problem exists for `\phi`.

ToruNiina:
Thank you for letting me know about the problem. I didn't care about it (actually, I automatically generated the lines).
I also want to see his opinion.

nyuichi:
Any news on this?

qnighy:
@gfngfn How do you think about the varepsilon problem? Do we introduce `\epsilon` for now with no compatibility guarantee? Do we only decide the ground default for now, and decide switching strategy (e.g. introducing an option to the document class or just provide two commands) later? We can also exclude the letters with variants for now.

qnighy:
I [asked](https://twitter.com/bd_gfngfn/status/987972346127433728) @gfngfn.

@gfngfn's opinion is to align with Unicode and have `\epsilonsymb` for U+1D716 (which has the same shape as TeX `\epsilon` when compared using [the reference shapes](https://unicode.org/charts/PDF/U1D400.pdf)).

I'm in favor of it, too. I suggest to merge this change and add `\epsilonsymb` later.

gfngfn:
Merged the PR. I’m extremely sorry to be too lazy… Thank you for sending the PR and holding the subsequent discussion.

Issue #26, opened 2018-02-12
consider using ocamlgraph for some graph structure implementation


objmagic:
seems `ocamlgraph`'s [directed graph](https://github.com/backtracking/ocamlgraph/blob/d48ae1a2977e65a32f6f7937a6a7975cbe7cf2b4/src/persistent.mli#L65) can be used to replace [src/backend/flowGraph.ml](https://github.com/gfngfn/SATySFi/blob/master/src/backend/flowGraph.ml)

PR #33, opened 2018-02-17
Faster Travis CI

nekketsuuu:
This PR makes CI faster:

|   | linux | osx | 
|:--|--:|--:|
| current `master` | 12--14 min | 13--15 min |
| this PR | 7--8 min | 13--15 min |

See [this result](https://travis-ci.org/nekketsuuu/SATySFi/builds/342762484) to confirm. Also, https://github.com/nekketsuuu/ocaml_travis_ci may be informative.

### Details

* For `linux`, I use a Docker image `ocaml/opam` in order to avoid compiling OCaml in each job.
* For `osx`, I directly download OPAM from the official repo, and then compile OCaml and SATySFi.

### Notes

* This PR (and also current `master`) doesn't run tests because currently we can't run tests for `batteries` with OCaml 4.06.0 (see ocaml-batteries-team/batteries-included#835).
* This PR also enables to build SATySFi with OCaml 4.05.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,15 +5,13 @@ services:
 matrix:
   include:
     - os: osx
-      env: OCAML_VERSION="4.06.1"
+      env: OCAML_VERSION="4.06.0"
     - os: linux
       env: DISTRO="alpine" OCAML_VERSION="4.05.0"
     - os: linux
       env: DISTRO="alpine" OCAML_VERSION="4.06.0"
     - os: linux
-      env: DISTRO="alpine" OCAML_VERSION="4.06.1"
-    - os: linux
-      env: DISTRO="ubuntu-16.04" OCAML_VERSION="4.06.1"
+      env: DISTRO="ubuntu-16.04" OCAML_VERSION="4.06.0"
 env:
   global:
     - PACKAGE="satysfi"
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,13 +5,15 @@ services:
 matrix:
   include:
     - os: osx
-      env: OCAML_VERSION="4.06.0"
+      env: OCAML_VERSION="4.06.1"
     - os: linux
       env: DISTRO="alpine" OCAML_VERSION="4.05.0"
     - os: linux
       env: DISTRO="alpine" OCAML_VERSION="4.06.0"
     - os: linux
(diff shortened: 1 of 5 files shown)


konn:
Just an idea, but how about using [caching mechanism](https://docs.travis-ci.com/user/caching/) to save more installation time?

nekketsuuu:
@konn Thanks for your suggestion! I've tried to cache `apt` and `$HOME/Library/Caches/Homebrew`, and I noticed there seems to be no effective differences (so I didn't implement it for simplicity).

However, it's interesting whether we can make a cache of OCaml compilers. I'll tackle this problem tonight. Note that just caching `~/.opam` may have a problem around installation of third-party packages.

nekketsuuu:
@konn It seems difficult to cache only compilers because both the compiler and packages are installed under `~/.opam/<OCaml version>/`. So I decide to make no cache in this PR. Anyway, thanks for letting me your idea :pray:

objmagic:
@nekketsuuu maybe we can try Docker based test

See [README](https://github.com/ocaml/ocaml-ci-scripts/blob/bebbd8ac917e9b0d8ab6e85a5ca9de5fe9a4902a/README-travis.md)

See examples:
1. https://github.com/mirage/mirage/blob/master/.travis.yml#L2
2. https://github.com/mirage/ocaml-cohttp/blob/master/.travis.yml

objmagic:
We can also shorten build time to make CI faster. For example, [switch to ocamlgraph and remove Core_kernel dependency](https://github.com/gfngfn/SATySFi/issues/26) will save us a lot of time since Core_kernel is still too heavy. It pulls in a lot of packages to install.

nekketsuuu:
@objmagic Thanks for your suggestion! Actually, I already implement Docker-based tests for linux. I also tried to use `.travis-docker.sh` of [`ocaml-ci-scripts`](https://github.com/ocaml/ocaml-ci-scripts), and knew the current script is faster than it for some reason. See https://github.com/nekketsuuu/ocaml_travis_ci for my other trials, or see [this comment](https://discuss.ocaml.org/t/what-are-the-best-practices-for-setting-up-continuous-integration-for-ocaml-projects/1499/3) for an other's method which I based on.

nekketsuuu:
This PR becomes a little old, so I close this for now.

PR #34, opened 2018-02-18
Support multiple configuration directories

nyuichi:
Satysfi's path resolver is modified as follows:
1. Satysfi first searches `~/.satysfi` for a config/package/font file the input saty program requires,
2. and if the search fails, satysfi then searches `/usr/local/share/satysfi`, and if it fails again, `/usr/share/satysfi` is searched.

Makefile now installs the initial configuration and fonts in the system directory by default. If you want use some extra files (say, Arno Pro fonts) in a satysfi code, you just need to make a directory `~/.satysfi/dist/fonts/` and put the font files there. (You don't even have to create empty directory trees under ~/.satysfi if you don't use them).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 PREFIX=/usr/local
-LIBDIR=~/.satysfi
+LIBDIR=$(PREFIX)/share/satysfi
 SRCROOT=src
 BACKEND=src/backend
 FRONTEND=src/frontend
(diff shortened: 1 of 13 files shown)


gfngfn:
Confirmed that the modified code works fine. Thank you!

gfngfn:
Oops, I mistakenly added another error message for `Config.DistFileNotFound(...)` insensible of you having written one… I deleted the duplicate error message.

Issue #35, opened 2018-02-19
autoconf is required to install SATySFi, autoconf is not preinstalled on macOS and Ubuntu


KilledByNLP:
Hello I love SATySFi so much. If we solve this problem, SATySFi must be much better!
# Problem
I tried to install SATySFi on macOS High Sierra(10.13.2), however `opam install satysfi` failed with below errors.
```
=-=- Processing actions -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=  🐫
[ERROR] The compilation of conf-autoconf failed at "which autoconf".

#=== ERROR while installing conf-autoconf.0.1 =================================#
# opam-version 1.2.2
# os           darwin
# command      which autoconf
# path         /Users/syrup/.opam/4.06.0/build/conf-autoconf.0.1
# compiler     4.06.0
# exit-code    1
# env-file     /Users/syrup/.opam/4.06.0/build/conf-autoconf.0.1/conf-autoconf-18608-ffb3fd.env
# stdout-file  /Users/syrup/.opam/4.06.0/build/conf-autoconf.0.1/conf-autoconf-18608-ffb3fd.out
# stderr-file  /Users/syrup/.opam/4.06.0/build/conf-autoconf.0.1/conf-autoconf-18608-ffb3fd.err



=-=- Error report -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=  🐫
The following actions were aborted
  ∗  install bitv    1.3
  ∗  install satysfi ~unknown
The following actions failed
  ∗  install conf-autoconf 0.1
No changes have been performed

=-=- conf-autoconf.0.1 troobleshooting =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=  🐫
=> This package relies on external (system) dependencies that may be missing. `opam depext
   conf-autoconf.0.1' may help you find the correct installation for your system.
```

# Solution
Just execute `brew install autoconf`

# Proposal
Add `brew install autoconf` in `README.md`

omasanori:
Note that it also applies to the Ubuntu instructions.

nekketsuuu:
Or, `opam depext -ui satysfi` resolves this. (I didn't write this because there is no `conf-zip` now.)

KilledByNLP:
Which is better?

nyuichi:
[...]

PR #36, opened 2018-02-19
Add autoconf in README.md to be required on macOS 

KilledByNLP:
Please see #35

### DIFF
--- a/README-ja.md
+++ b/README-ja.md
@@ -15,6 +15,7 @@
 
 ビルド前に最低限，以下のソフトウェアが必要です。
 
+* autoconf
 * bzip2
 * cc
 * git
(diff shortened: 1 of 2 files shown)


nekketsuuu:
How about add `autoconf` into not only macOS, but also Ubuntu? That is, add `autoconf` to the list of required softwares, and add `autoconf` into the line of `apt-get install`.

KilledByNLP:
Thank you for your kind advice, that's right!
I added it.

nekketsuuu:
@KilledByNLP Could you add `autoconf` on the top of `Prerequisites`' list? Also, how about sorting packages alphabetically (i.e., `brew install autoconf git opam unzip wget`)?

KilledByNLP:
@omasanori @nekketsuuu Thanks! All is done!

KilledByNLP:
@omasanori @gfngfn Conflict is resolved

omasanori:
@gfngfn ping?

qnighy:
The indirect dependency on `conf-autoconf` through `bitv` is now removed by #45. I'm not sure if we still need autoconf.

PR #41, opened 2018-02-24
Remove unit token.

qnighy:
Remove `()` token and parse `(` `)` as the unit. I think this is more consistent within SATySFi and with other ML-like languages.


qnighy:
I'm going to prepare a more comprehensive collection of patches instead.

PR #42, opened 2018-02-24
Remove redundant dependencies for macOS

equal-l2:
`unzip` is bundled with macOS.
`git` is provided by Xcode Command Line Tools, which is a dependency of
Homebrew.
We don't have to install them via Homebrew, unless we need the latest version of them.

### DIFF
--- a/README-ja.md
+++ b/README-ja.md
@@ -47,7 +47,7 @@ opam update
 # また，Homebrew もインストールしてください。
 
 brew update
-brew install unzip wget git opam
+brew install wget opam
 
 # 以下のコマンドは OPAM が（~/.bash_profile などの）ファイルに環境変数に関する設定を追記してもよいか聞いてきます。
 # 必ず説明を読み，環境変数を適切に設定してください。
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ opam update
 # Also, install Homebrew.
 
 brew update
-brew install unzip wget git opam
+brew install wget opam
 
 # The following command will ask if OPAM modifies some files.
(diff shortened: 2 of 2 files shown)


nekketsuuu:
Did you check it's bundled with older OS of Mac? I myself wrote that line without knowing which one is installed at default because I don't have any Mac. I only knew that GCC and Make are installed by Xcode. Could you tell me how to look for bundled softwares if you know?

equal-l2:
Installing `unzip` via Homebrew doesn't solve the problem because `unzip` is a *keg-only*, which means the software won't be installed in the usual place (e.g. `/usr/local`).
By the way, `unzip` is a *keg-only* because `macOS already provides this software and installing another version in parallel can cause all kinds of trouble.` according to Homebrew, so we can assume `unzip` is bundled with any version of macOS supported by Homebrew, macOS 10.11, at least.

`git` is contained even in the oldest Command Line Tools, which have released for Mac OS 10.7 in 2012.

equal-l2:
Excerpt from Homebrew/homebrew-dupes@4fc048c (first commit of Homebrew's `unzip`):
>OS X <=10.8 have unzip < 6, which don't support ZIP64 files

nekketsuuu:
@equal-l2 Thank you for your reference! It's really helpful. Then I agree to merge this change. Thanks :)

equal-l2:
Is there any blocking issue, @nekketsuuu?

nekketsuuu:
@equal-l2 I think no. But only @gfngfn has the merge permission.

gfngfn:
I am so sorry for the late response… Thank you for fixing READMEs.

PR #43, opened 2018-02-28
A collection of refactoring/bugfixes to the lexer

qnighy:
This PR is a collection of refactoring and bugfixes to the lexer, especially reducing statefulness of the lexer to the minimum degree.


gfngfn:
Confirmed that the refactored frontend works fine. Thank you very much for the great fixes.

PR #45, opened 2018-03-02
Remove uCoreLib from dependencies

### LINKED ISSUE #40: Switch from uCoreLib to uucp and utext
Daniel just released his another Unicode library: https://github.com/dbuenzli/utext

As a side note, I recommend using packages used by [Mirage organization](https://github.com/mirage) or packages developed by people from OCaml Labs since those packages are **well documented** and **well maintained**

qnighy:
SATySFi does not use much of uCoreLib. This PR replaces the usages of uCoreLib with batteries and uutf, which it already depends on.

Fixes #40.

#35 may also be related.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,6 +4,3 @@
 [submodule "backend/external/otfm"]
 	path = external/otfm
 	url = https://github.com/gfngfn/otfm.git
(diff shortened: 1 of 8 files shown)


qnighy:
Rebased.

PR #48, opened 2018-03-03
Implements `break-on-regex` and `string-match` primitives

konn:
This pull-request implements `break-on-regex` and `string-match` primitives, which build on top of `Str` package.
These are useful to implement a lightweight syntax highlighting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ OCB_FLAGS = -cflags -unsafe-string -cflag -w -cflag -3 \
 	-I $(SRCROOT)/ -I $(FRONTEND)/ -I $(BACKEND)/ -I $(CHARDECODER)/ \
 	-I $(EXTERNAL)/otfm/src/ -I $(EXTERNAL)/camlpdf/ \
 	-I $(EXTERNAL)/ucorelib/src/ \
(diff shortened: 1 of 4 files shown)


qnighy:
I'm not totally against this PR, but we should at least note that it exposes [regex syntax of `Str`](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Str.html). It's fortunate that the syntax is fairly standard and simple, therefore does not prevent portability much.

konn:
@qnighy Exactly. I chose `Str` just because it is a part of standard library and no additional package dependency is required. If we once agree to add good package as a dependency, I will rewrite on it.

gfngfn:
Thank you very much for sending an ambitious PR.

I extended your modification by making a distinction between types `string` and `regexp` for convenience and clarity. The construction of a regular expression from a string can be performed by a newly introduced primitive `regexp-of-string : string -> regexp`. Therefore, the package [Steamer](https://github.com/konn/satysfi-steamer/blob/master/steamer.satyh) of `konn`-san’s own making needs a slight correction in order to keep its functionality.

konn:
Thanks!

PR #52, opened 2018-03-03
mention homebrew formula

nyuichi:


gfngfn:
Thank you for enhancing READMEs.

PR #53, opened 2018-03-03
reduce prerequisite (use curl if possible)

nyuichi:


nyuichi:
Travis (for Mac) failed for some reasons, but the patch itself should be fine :(

Makefile
```diff
@@ -47,9 +47,14 @@ install: $(TARGET)
 lib:
 # -- downloads fonts --
 	mkdir -p temp/
-	wget -N http://www.gust.org.pl/projects/e-foundry/latin-modern/download/lm2.004otf.zip -P temp/
+	if [ -x "$$(command -v curl)" ]; then \
```
KeenS: Only `if command -v curl > /dev/null` will do because `command` searches only executables.
Moreover, this form rejects aliases, shell functions or like those but simple `command -v` returns true for such command like stuff.

nyuichi: Ah, yes. you’re right, I missed that point. Do you have any workaround for this?

nyuichi: I am not even sure if we should support such a case.
It seems that people usually use aliases and functions only in interactive shells [1][2].

- [1] https://wiki.archlinux.org/index.php/bash#Aliases
- [2] https://wiki.archlinux.org/index.php/Bash/Functions

KeenS: > Do you have any workaround for this?

Use `if command -v curl > /dev/null`.

> I am not even sure if we should support such a case.

I'm sorry, I was just a foolish. This was a mere coding style problem. Please code in your style.

nyuichi:
rebased.

gfngfn:
@nyuichi @equal-l2 I’m terribly sorry for the extremely late response. Thank you.

nyuichi:
@gfngfn 

No problem! Please don't say sorry, we don't hurry you. Anyway, thank you so much for developing & maintaining this great project!

PR #55, opened 2018-03-05
Unicode-aware string primitives

konn:
This pull-request makes `string-sub` and `string-length` compatible with Unicode characters.

* Renames `string-sub` and `string-length` to `string-sub-bytes` and `string-byte-length`.
* Re-implements `string-sub` and `string-length` using BatUTF8 operations.


qnighy:
@gfngfn does anything block this from merging?

PR #56, opened 2018-03-05
Implements `string_scan`

konn:
Implements `string_scan` primitive to scan string by regexp.
This is still using `Str` module, but it might be better to switch to other library, because `Str` seems incapable of treating url correctly.
For example, `Str.split (regexp "") "あいう"` splits every bytes apart, not unicode char.


gfngfn:
I am so sorry to let you wait for a long time. `Str` seems obsolescent, but since it is meaningful to provide `string-scan` as a primitive, I merged the PR for now. We can, of course, refactor the implementation of the primitive if needed in the future. Anyway, thank you for the improvement!

PR #58, opened 2018-03-06
use Junicode fonts in stdja.satyh

pandaman64:
Even though 042a8aa replaced ArnoPro with Junicode in `stdjabook.satyh`, `stdja.satyh` still uses ArnoPro as its roman font.
This makes several files (such as ones in `doc` directory) uncompilable without ArnoPro.
This commit fixes the issue by applying the same kind of patch as 042a8aa to `satja.satyh`.


gfngfn:
I’m sorry for the late response. Thanks for amending!

PR #60, opened 2018-03-06
Implements `string-compare`

konn:
Implemented `string-compare` primitive, which corresponds to the OCaml's `String.compare` function:
 `string-compare l r` returns `0` if they are equal, negative value if `l` is strictly smaller
 (w.r.t. lex-ordering) than `r`, positive if greater.

I once considered to add a special type dedicated for comparison result, e.g. Haskell's `Ordered`-type,
but I didn't take that way because it introduces additional type for less used things.

### DIFF
--- a/src/frontend/evaluator.ml
+++ b/src/frontend/evaluator.ml
@@ -1319,6 +1319,10 @@ and interpret env ast =
       let str2 = interpret_string env ast2 in
         BooleanConstant(String.equal str1 str2)
 
+  | PrimitiveStringCompare(ast1, ast2) ->
+      let str1 = interpret_string env ast1 in
+      let str2 = interpret_string env ast2 in
+        IntegerConstant(String.compare str1 str2)
 
   | PrimitiveStringSub(aststr, astpos, astwid) ->
       let str = interpret_string env aststr in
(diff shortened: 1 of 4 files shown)


gfngfn:
Thank you for suggestion! I consider that introducing `string-explode : string -> int list`, a primitive operation more general than `char-code`, may be a better choice:

    string-explode `ABC` %---> [65; 66; 66]

    let char-code s =
      match string-explode s with
      | i :: _ -> i

In addition it may be a better choice to prepare a new base type `uchar` for Unicode code points and to introduce

* `string-explode : string -> uchar list`
* `uchar-code : uchar -> int`
* `get-uchar : int -> uchar option`

etc.

puripuri2100:
Added `string-explode: string -> int list` by #202 .
So, `string-compapre` can be implemented on SATySFi.

ex:
```
let string-compare sl sr =
    let lstl = string-explode sl in
    let lstr = string-explode sr in
    let-rec sub lstl lstr =
        match (lstl, lstr) with
        | ([],[]) -> 0
        | (_, []) -> 1
        | ([],_) -> -1
        | (nl::xl,nr::xr) -> (
            if nl < nr then
                -1
            else
                if nl > nr then
                    1
                else
                    sub xl xr
        )
    in
    sub lstl lstr
```

Issue #14, opened 2018-02-09
Use opam repo instead of submodules


objmagic:
See: https://github.com/ocamllabs/advanced-fp-repo

A customized repo can be set up to have your customized camlpdf, otfm, and ucorelib there.

I can send some pull requests when I'm free.

nekketsuuu:
@gfngfn Will you plan to send pull requests of your custom packages? If not, I also think creating a opam repo is a good idea to divide the dependency problem and to simplify the build process. But it needs a new official git repo, so now I can't contribute to this.

cf.: https://opam.ocaml.org/doc/Usage.html#opam-repo

objmagic:
@nekketsuuu you can create a opam repo, test it, and transfer ownership back to gfn on GitHub. Or better, we can set up an organization on GitHub.

nekketsuuu:
@objmagic Thanks! I don't know we can transfer repo's ownership.

nekketsuuu:
@gfngfn Which do you think is better, to transfer ownership or to make an organization?

qnighy:
So in my understanding, the possible options are:

1. Keep status quo.
2. Create a custom opam repo.
3. Try to push everything back to the upstreams, then use the official package.

We can also try to push *some* of the patches back to the upstreams, while choosing 1 and 2.

@gfngfn which option do you think is ideal? My idea is that if these patches are highly SATySFi-specific, then 2 is suitable. I'm willing to help you either way.

gfngfn:
Thank you for the suggestion, and I’m terribly sorry for the late response.

For now, I would not like to send `gfngfn/otfm:master` to `dbuenzli/otfm` as a pull request, Since the development of `gfngfn/otfm` largely breaks the backward compatibility with `dbuenzli/otfm`. If willing to do so, it seems necessary to have a profound discussion and to reach a genuine consensus with `dbuenzli`-san.

nekketsuuu:
I transferred the ownership of nekketsuuu/satysfi-external-repo to gfngfn.

nekketsuuu:
@objmagic #62 is merged thanks to gfn-san :-)

gfngfn:
Thank you very much for introducing external opam repositories (and am very sorry for the late response).

objmagic:
Great. Thanks!

objmagic:
closed via #62

PR #62, opened 2018-03-14
Use opam repo

### LINKED ISSUE #23: Missing dependency of uchar
I noticed that a package [`uchar`](https://opam.ocaml.org/packages/uchar/) is not written in [`opam`](https://github.com/gfngfn/otfm/blob/8f1086ea1773d665cf8e24ecd6c84551e0cc94ed/opam#L16) whereas it's needed for the external package `otfm` ([ref](https://github.com/gfngfn/otfm/blob/8f1086ea1773d665cf8e24ecd6c84551e0cc94ed/opam#L16)). Also, a constraint `"uutf" {>= "1.0.0"}` is missing.

Although `uchar` is unnecessary for OCaml >= 4.03, this should be written in `opam`.
Or, using opam repo (#14) will help.

nekketsuuu:
Implements #14. Also fixes #23.

cc: @nyuichi, if merged, this PR will break the Homebrew formula.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,0 @@
-[submodule "external/camlpdf"]
-	path = external/camlpdf
-	url = https://github.com/gfngfn/camlpdf
-[submodule "backend/external/otfm"]
-	path = external/otfm
-	url = https://github.com/gfngfn/otfm.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ install:
   - wget https://raw.githubusercontent.com/ocaml/ocaml-travisci-skeleton/master/.travis-ocaml.sh
   - bash -ex .travis-ocaml.sh
   - eval `opam config env`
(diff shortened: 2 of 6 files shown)


nyuichi:
@nekketsuuu 

I will fix the formula. BTW, what is the motivation of this pull request?

objmagic:
LGTM!

@nyuichi because using external opam repo is a better solution than git submodules. Current approach of using git submodules looks quite hacky to me...

nekketsuuu:
@nyuichi using the repo simplifies the dependency problem (#23) and the build script of SATySFi itself. Vendoring (i.e., copying libraries under the project root directory) is easy, but sometimes triggers a complex behavior. For example, there may be a conflict if we installed the original version of otfm via OPAM and want to build SATySFi with the vendored version of otfm. So I think only using OPAM is better than making git submodules in order to manage libraries.

nyuichi:
@nekketsuuu 

Thank you for the explanation. That makes sense to me.

Issue #46, opened 2018-03-02
Travis started to fail


qnighy:
Currently #44 ([build #929](https://travis-ci.org/gfngfn/SATySFi/builds/347738475)) and #45 ([build #930](https://travis-ci.org/gfngfn/SATySFi/builds/348081919)) fail (for seemingly the same reasons). As #44 is a trivial change, they must be spurious errors.

I couldn't figure out the exact cause. Seems to be related to `core_kernel`.

mt-caret:
I get the same error message as https://travis-ci.org/gfngfn/SATySFi/builds/348081919 while packaging SATySFi for Nix; I would love to know how to fix it.

Logs:
https://gist.github.com/mt-caret/26689bcc94a81274081ff3927fa1d9c5

I'm at 042a8aaf3b61785b8edac7d799067ad758ed2fd7 and I'm getting this error since 26aa766d49ddc297741d72ddb2ba8d40457cc61c so the error might be originating somewhere else...

qnighy:
I tried to bisect `dune` (formerly `jbuilder`) between `1.0+beta17` and `1.0+beta18` with the following test script:

```sh
opam reinstall -y jbuilder
opam reinstall -y core_kernel
```

Bisect reported ocaml/dune@9e4cc843704262b3cb563096ae7df128d0ea2bff as the first bad commit.

mt-caret:
I can confirm, https://github.com/gfngfn/SATySFi/pull/47 indeed fixes the error. Thank you!

An upstream bug, then?

qnighy:
Probably. I'm investigating further.

qnighy:
The bisect procedure above was wrong; now I pinned `jbuilder` and `satysfi` to local repos and bisected on `jbuilder` with the following script:

```sh
opam reinstall -y jbuilder
opam reinstall -y core_kernel
opam reinstall -y satysfi
```

Bisect reported ocaml/dune@b3838284c6f28aaa0afabf865ce126c20e2f3616 as the first bad commit.

mt-caret:
There's a PR that might be addressing the issue: https://github.com/ocaml/dune/issues/567

qnighy:
Confirmed locally that the current master (ocaml/dune@e4aac2da97c069a914c9fb6c06697b5a99145d17) does not cause the `Core_kernel__` problem. I think it is good to pin to `1.0+beta17` until `1.0+beta19` is released.

PR #65, opened 2018-03-15
Remove jbuilder pin as ocaml/dune#567 is resolved

qnighy:
This commit removes part of version pinnings introduced by bd737ab and f8189a8, because jbuilder 1.0+beta18.1 includes a fix of ocaml/dune#567. Part of #46.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,6 @@ install:
   - bash -ex .travis-ocaml.sh
   - eval `opam config env`
   - opam update
-# Issue #46: we temporarily pin jbuilder to avoid problems in 1.0+beta18.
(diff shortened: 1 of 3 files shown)


sakas--:
@qnighy Does this commit allow us to unpin camlimages?

qnighy:
@sakas-- They're orthogonal. I've not confirmed whether we still need camlimages pin.

qnighy:
Note that camlspotter gracefully resolved the Graphics issue immediately after my report on Twitter. However, when I tried to unpin camlimages after that, a mysterious hash mismatch occured on Travis. Not sure about the current state of travis.

Issue #22, opened 2018-02-11
Set version and release


nekketsuuu:
Currently, there is no version number of SATySFi.
How about giving it a (semantic) version number?

Also, CHANGELOG would be needed.

nekketsuuu:
SATySFi v0.0.1 is [(pre-)released](https://github.com/gfngfn/SATySFi/releases/tag/v0.0.1)! Thanks gfn-san :+1:

PR #66, opened 2018-03-16
Add Nix installation method

mt-caret:
Nix supports macOS too (and Bash on Ubuntu on Windows unofficially), but I've only tested the build on NixOS and Ubuntu. (If I have the time) I'll try them and if they seem to work fine, I'll add them to the supported list.


nekketsuuu:
Why is it better to support installation by Nix officially? I think Nix is not so popular.

mt-caret:
Considering that getting SATySFi packaged into distros would probably be a pain, wouldn't it be nice for there to be at least one way to install SATySFi on Linux distributions (which doesn't require you to manually build it from source)?

nekketsuuu:
Hmm... I think packaging SATySFi is early because it's under the highly development period currently. There is even no version number (#22). So I think the cost to support Nix officially is bigger than the advantage of that.

Note that this is my personal opinion. The final decision will be done by gfn-san.

README.md
```diff
@@ -8,6 +8,14 @@
 
 This software is supported by IPA Mitou Project 2017 (see the abstract [here](https://www.ipa.go.jp/jinzai/mitou/2017/gaiyou_t-4.html)).
 
+## Install using Nix (for Linux users)
+
+Follow the instructions here [here](https://nixos.org/nix/download.html) to install the Nix package manager.
```
nekketsuuu: Typo here

qnighy:
> Why is it better to support installation by Nix officially? I think Nix is not so popular.

I don't think of README inclusion as an indication of official support. However, if we want to be clear, I suggest adding the "list of third-party packages/tools" section.

qnighy:
As a side note, I also want some traffic line from here to [satysfi.vim](https://github.com/qnighy/satysfi.vim) and [SATySFi for Windows](https://github.com/qnighy/satysfi-cross-windows), whether "official" or not.

qnighy:
@mt-caret I asked @gfngfn. His [opinion](https://twitter.com/bd_gfngfn/status/995182589005611008) is that it should be clear that the Nix package is unofficial.

mt-caret:
HEAD of master currently doesn't build on Nix, so I'll rebase and add something saying it's unofficial when I update the package over at NixOS/nixpkgs.

mt-caret:
Considering that SATySFi has had semver for a while and [you can install 0.0.4 via nixpkgs-unstable now](https://repology.org/project/satysfi/versions), what does everybody think about merging this?

Issue #38, opened 2018-02-19
lib-satysfi is copied to unexpected directory when `opam install satysfi`


nekketsuuu:
## Problem

It seems that `opam install satysfi` doesn't copy `lib-satysfi` to an appropriate place.

```sh
# After `opam install satysfi`...
SATySFi/tests$ make
satysfi list.satyh math.satyh head.satyh first.saty -o output.pdf
 ---- ---- ---- ----
  target file: 'output.pdf'
! [Error] package file not found: dist/hash/default-font.satysfi-hash
Makefile:4: recipe for target 'all' failed
make: *** [all] Error 1
```

## Reason

This problem is related to a breaking change requested by #34. This PR changes the default location of the library to `/usr/local/share/satysfi`, but OPAM copies it under `$HOME/.opam/4.06.0/share/satysfi` because `PREFIX=%prefix%` while running `opam install satysfi`.

And, we shouldn't copy libraries under `/usr/local/share` while `opam install`, because [the OPAM manual](https://opam.ocaml.org/doc/Manual.html#opamfield-install) says:

> *(the installation commands)* should only write to subdirectories of `prefix`, without altering the source directory itself.

In fact, copying to `/usr/local/share` requires `sudo` on Ubuntu.

## Proposal

This problem will occur if non-default `$PREFIX` is set. <del>How about making a config file which has all library path? I'll make a PR of this.</del>

No, this proposal is the same as using `~/.satysfi`. This causes a problem for Homebrew. Some other solution is needed (for example, using an environment variable like `$LD_LIBRARY_PATH`?).

## Workaround

To newcomers: you can fix this issue temporarily by `cp -R "$(opam config var prefix)/share/satysfi" "${HOME}/.satysfi"`.

## Environment

* Ubuntu 17.10
* OCaml 4.06.0, OPAM 1.2.2
* SATySFi 3e6d067a7aecb4a8d986831a50810fc2ac71fe19

omasanori:
How about just making SATySFi `$PREFIX`-friendly? (At least `setup_root_dirs` will need some changes.) Maybe I am missing the point, though...
`PATH`-like environment variable would be helpful anyway.

nekketsuuu:
Survey: TeX uses multiple `TEXMF`.

References (in Japanese):

[...]

PR #67, opened 2018-03-18
Change root dirs in Win32.

qnighy:
This patch changes root dirs in Win32, because on Windows,
- `%HOME%` is rarely used. `%userprofile%` is set instead.
- Installation directories are movable.

After this patch, SATySFi [sets the root dirs](https://twitter.com/qnighy/status/974919889013022720) to
- `%userprofile%\.satysfi;%SATYSFI_RUNTIME%` in Win32.
- `$HOME/.satysfi:/usr/local/share/satysfi:/usr/share/satysfi` in other
  platforms.

### DIFF
--- a/src/frontend/main.ml
+++ b/src/frontend/main.ml
@@ -729,10 +729,25 @@ let arg_spec_list curdir =
 
 
 let setup_root_dirs () =
-  let ds =
-    (match Sys.getenv_opt "HOME" with
-    | None    -> []
-    | Some(s) -> [Filename.concat s ".satysfi"]) @ ["/usr/local/share/satysfi"; "/usr/share/satysfi"] in
+  let runtime_dirs =
+    if Sys.os_type = "Win32" then
+      match Sys.getenv_opt "SATYSFI_RUNTIME" with
+      | None    -> []
+      | Some(s) -> [s]
+    else
+      ["/usr/local/share/satysfi"; "/usr/share/satysfi"]
+  in
+  let user_dirs =
+    if Sys.os_type = "Win32" then
+      match Sys.getenv_opt "userprofile" with
+      | None    -> []
+      | Some(s) -> [Filename.concat s ".satysfi"]
+    else
+      match Sys.getenv_opt "HOME" with
+      | None    -> []
+      | Some(s) -> [Filename.concat s ".satysfi"]
+  in
+  let ds = user_dirs @ runtime_dirs in
   (if List.length ds = 0 then
     raise NoLibraryRootDesignation);
   Config.initialize ds


nekketsuuu:
Are there any reasons to use `SATYSFI_RUNTIME` only in Win32? In #38, nyuichi proposed [an idea](https://github.com/gfngfn/SATySFi/issues/38#issuecomment-366835572) to use variables on all systems.

Also, note that a similar environment variable `SATYSFI_LIB_ROOT` was used in the past. That is deprecated after aa03c8201fac3bb62f11d2f07cb3c53a04e4baab.

qnighy:
> Are there any reasons to use `SATYSFI_RUNTIME` only in Win32? In #38, nyuichi proposed an idea to use a variable on all systems.

Not actually. I was just conservative with other platforms. Let's discuss some now, but I think it's good to defer the final decision until we get some feedbacks.

Possible discussions are:

- Should an environment variable be used on other platforms? I think yes.
- Should the fixed paths (`/usr/[local/]/share/satysfi`) be removed? I think no for now.
- Name of the environment variable? It's a bikeshed. I picked `*_RUNTIME` from vim, but `SATYSFI_LIB_ROOT` also seems reasonable. Which do you like?

nekketsuuu:
> Not actually. I was just conservative with other platforms. Let's discuss some now, but I think it's good to defer the final decision until we get some feedbacks.

Then we can just postpone the discussions since this PR is only for the Windows support. I can create another PR to support the environment variable or something. How about removing `%SATYSFI_RUNTIME%` in this PR if it's unnecessary now?

qnighy:
It's necessary. Please see the link above. I'm not in hurry to merge only part of it, because the whole patch is already applied in qnighy/satysfi-cross-windows.

nekketsuuu:
So [you want](https://twitter.com/qnighy/status/974913342199414784) to change search paths depending on an installed location, right? That makes sense.

Then I agree to [your previous comment](https://github.com/gfngfn/SATySFi/pull/67#issuecomment-373971220). We should use an environment variable on all platforms. Any variable name seems to be OK to me (yes, this is a bikeshed). I personally prefer `SATYSFI_LIB_ROOT`, with no strong reason. Since `SATYSFI_RUNTIME` is also good enough, I think there is no need to change.

To summarize, I agree to:

* Use `SATYSFI_RUNTIME` on Windows and other distros.
* Leave the fixed search path (`/usr/[local/]/share/satysfi`) currently.

Any opinions?

qnighy:
> So you want to change search paths depending on an installed location, right? That makes sense.

Yes, windows binary distributions are usually assumed to be "relocatable" (in the sense of file paths).

I'll wait for opinions and then change `SATYSFI_RUNTIME` to `SATYSFI_LIB_ROOT` if it's more favored.

gfngfn:
I merged your pull request for the time being. If the original name `SATYSFI_LIB_ROOT` is preferable to `SATYSFI_RUNTIME` in the long run (e.g. for releases), we can modify it with ease.

Thank you.

PR #69, opened 2018-03-30
Say crossref resolved only when no `?`.

### LINKED ISSUE #68: SATySFi says that all cross references were solved while they aren't
Consider the following `mwe.saty`. Note that ``\ref(`sec:fooo`);`` is a typo on purpose.

```
@require: stdja

StdJa.document (|
  title = {};
  author = {};
  show-title = false;
  show-toc = false
|) '<
  +section ?:(`sec:foo`) {Foo} <
    +p {
      This is Section \ref(`sec:fooo`);.
    }
  >
>
```

If we typeset this document, we obtain the log shown below and it claims `all cross references were solved.` after the 2nd trial.

```console
$ satysfi mwe.saty -o mwe.pdf
 ---- ---- ---- ----
  target file: 'mwe.pdf'
  dump file: 'mwe.satysfi-aux' (will be created)
  parsing 'mwe.saty' ...
  parsing 'stdja.satyh' ...
  parsing 'pervasives.satyh' ...
  parsing 'gr.satyh' ...
  parsing 'list.satyh' ...
  parsing 'math.satyh' ...
  parsing 'color.satyh' ...
 ---- ---- ---- ----
  reading 'pervasives.satyh' ...
  type check passed.
 ---- ---- ---- ----
  reading 'list.satyh' ...
  type check passed.
 ---- ---- ---- ----
  reading 'color.satyh' ...
  type check passed.
 ---- ---- ---- ----
  reading 'math.satyh' ...
  type check passed.
 ---- ---- ---- ----
  reading 'gr.satyh' ...
  type check passed.
 ---- ---- ---- ----
  reading 'stdja.satyh' ...
  type check passed.
 ---- ---- ---- ----
  reading 'mwe.saty' ...
  type check: document
 ---- ---- ---- ----
  evaluating texts ...
  evaluation done.
 ---- ---- ---- ----
  breaking contents into pages ...
  needs another trial for solving cross references...
 ---- ---- ---- ----
  evaluating texts (2nd trial) ...
  evaluation done.
 ---- ---- ---- ----
  breaking contents into pages ...
  all cross references were solved.
 ---- ---- ---- ----
  embedding fonts ...
 ---- ---- ---- ----
  writing pages ...
 ---- ---- ---- ----
  output written on 'mwe.pdf'.
```

However, SATySFi obviously cannot find a label entry for `sec:fooo` and a question mark is printed on the output PDF. This situation wouldn't be considered as "solved" IMHO.

qnighy:
Fixes #68.

@gfngfn could you review this?

### DIFF
--- a/src/frontend/logging.ml
+++ b/src/frontend/logging.ml
@@ -74,7 +74,7 @@ let achieve_fixpoint unresolved_crossrefs =
   if unresolved_crossrefs = [] then
     print_endline ("  all cross references were solved.")
   else
-    print_endline ("  some cross references were not solved.")
+    print_endline ("  some cross references were not solved: " ^ String.concat " " unresolved_crossrefs ^ ".")
 
 
 let end_output file_name_out =
--- a/src/frontend/logging.ml
+++ b/src/frontend/logging.ml
@@ -67,11 +67,14 @@ let needs_another_trial () =
 
 
 let achieve_count_max () =
-  print_endline ("  some cross references were not solved.")
+  print_endline ("  could not reach to fixpoint when resolving cross references.")
 
 
-let achieve_fixpoint () =
-  print_endline ("  all cross references were solved.")
+let achieve_fixpoint unresolved_crossrefs =
+  if unresolved_crossrefs = [] then
+    print_endline ("  all cross references were solved.")
+  else
+    print_endline ("  some cross references were not solved.")
 
 
 let end_output file_name_out =
--- a/src/backend/crossRef.ml
+++ b/src/backend/crossRef.ml
@@ -114,6 +114,12 @@ let register (key : string) (value : string) =
         end
 
 
+let probe (key : string) =
+  match CrossRefHashTable.find_opt main_hash_table key with
+  | Some(value) -> Some(value)
+  | None -> None
+
+
 let get (key : string) =
   match CrossRefHashTable.find_opt main_hash_table key with
   | Some(value) -> Some(value)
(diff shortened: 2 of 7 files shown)


src/frontend/logging.ml
```diff
@@ -67,11 +67,14 @@ let needs_another_trial () =
...
+  print_endline ("  could not reach to fixpoint when resolving cross references.")
 
 
-let achieve_fixpoint () =
-  print_endline ("  all cross references were solved.")
+let achieve_fixpoint unresolved_crossrefs =
+  if unresolved_crossrefs = [] then
+    print_endline ("  all cross references were solved.")
+  else
+    print_endline ("  some cross references were not solved.")
```
sakas--: Why don't you print `unresolved_crossrefs`?
I think it's informative.

qnighy: I just threw away problems like formatting and sanitization. Let's implement it simply.

qnighy:
@gfngfn Thank you for reviewing! I implemented `probe-cross-reference` that does not record crossref resolution failure. Yet I'm not really sure the concrete situation where we need `probe-cross-reference`. Branching over crossref resolution status seems weird.

For logging part, in my opinion, logging internal things is better than logging nothing. There's still a room for improvement by allowing custom missing-cross-reference messages.

sakas--:
I suggested logging keys because I thought it's better than logging nothing.
In my opinion, reporting elaborated messages or adding a verbose mode can be addressed by another PR.
Maybe, we should prepare a place to discuss the overall design of error reporting/logging since providing "nice error messages" should be a strong point of SATySFi.

qnighy:
Any news on this? I'm ok with or without `probe-cross-reference`.

PR #70, opened 2018-04-04
Handle JPEG images with different color models

### LINKED ISSUE #31: Inconsistency about the `/ColorSpace` entry for embedded JPEG images
https://twitter.com/kurgm/status/963657925402165249

sakas--:
Fixes #31

Additional comments:
- The update on dune seems to have solved the issue on building CamlImages.  At least, the travis worked when I pushed to my fork repo.

- SATySFi can now load a jpeg image with CMYK color mode, but it cannot output a proper image.  
  (It outputs an image with wrong (negated?) colors,  but this behaviour is better than before; before the image was losing its shape)  
  As a temporary solution, I added a warning that notifies the fact that one is using an image with CMYK color model.  
  I think I know why SATysFi cannot produce a proper image.  It's because SATySFi doesn't embed a /Decode array into the pdf. (I can open a separate issue for this)

- I included a simple test file.  
  I want someone to check it before this gets merged since I only tested with my environment (Arch Linux).   I'd appreciate it if one could test other images.
  I don't know if we should include the test file to the final version of this PR.  I can remove it if we don't want to add extra files.


gfngfn:
Confirmed that it works (except for CMYK images). Thank you very much.

PR #72, opened 2018-05-07
add demo/satysfi-logo.jpg to generate demo.pdf

e-kwsm:
At present `demo/` does not have `satysfi-logo.jpg`, so `make` fails.
In this PR I use symlink to `tests/images/satysfi-logo-rgb.jpg` (possibly there is more preferable one).

### DIFF
--- a/demo/satysfi-logo.jpg
+++ b/demo/satysfi-logo.jpg
@@ -1,0 +1,1 @@
+../tests/images/satysfi-logo-rgb.jpg


e-kwsm:
Currently `logo/` has only `macrodown.svg` and I think the logo of SATySFi should also be located.
A symlink to it is natural.

nekketsuuu:
Indeed. Is it OK @gfngfn ?

PR #73, opened 2018-05-14
add new math commands

forestaa:
I added new math commands needed to write my report.
They are basically implemented by math-char, but \abs and \norm are implemented by math-paren.

### DIFF
--- a/lib-satysfi/dist/packages/math.satyh
+++ b/lib-satysfi/dist/packages/math.satyh
@@ -45,6 +45,8 @@ module Math : sig
   direct \times : [] math-cmd
   direct \nabla : [] math-cmd
 
+  direct \le     : [] math-cmd
+  direct \ge     : [] math-cmd
   direct \equiv  : [] math-cmd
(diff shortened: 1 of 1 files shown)


nekketsuuu:
Nice work! Maybe it's a little controversial that `\rightarrow` is mapped to `⇒` in your implementation, whereas in LaTeX, `\rightarrow` is `→` and `\Rightarrow` is `⇒`. The same thing can be said for `\leftarrow` and `\leftrightarrow`.

forestaa:
Thank you for your reviewing.
I think I should follow the latex's manner.

gfngfn:
Thank you for sending a nice PR! I merged it but modified `\le` to `\leq`, since `\leq` seems more popular than `\le` (and similarly for `\ge` and `\geq`). Though we could provide both `\le` and `\leq` according to the convention of LaTeX, for now I think it is somewhat awkward to provide more than one command names that indicate the same meaning.

PR #74, opened 2018-05-15
simplify `opam install` invocation in .travis.yml

leque:
All dependencies are already listed in the `opam` file. There should be no need to repeat them in `.travis.yml`.


gfngfn:
Merged. Thank you for sending a PR! (and sorry for the late response)

Issue #63, opened 2018-03-14
Interpreter Optimization


qnighy:
# Background

SATySFi interpreter is implemented as a [naïve AST traversal](https://github.com/gfngfn/SATySFi/blob/508b403e402fc464210011859f1ad13b40b22bb8/src/frontend/evaluator.ml#L470) now. This is typically simple but slow. Now, contrary to [expectations](https://twitter.com/bd_gfngfn/status/970595465166073856), we can find SATySFi documents doing complex computations such as [peg](https://github.com/konn/satysfi-steamer/blob/6fe84da5802161db4e833072b858a1e2cd462af5/peg.satyh), [pi](https://twitter.com/zr_tex8r/status/973842520965656576), and [mandelbrot](https://twitter.com/zr_tex8r/status/973538430427656194). They can also suffer from stack overflow (see #57), not only slowness.

Therefore, [it's a good time](https://twitter.com/bd_gfngfn/status/973849857092812800) to make SATySFi faster.

# How to do it

Typically, the next step is to introduce a VM. Other options would be to introduce JIT or full compilation, but I don't think they're worth doing now.

PR #77, opened 2018-05-27
add bytecode compiler and vm

matsud224:
This pull-request adds a bytecode compiler and a virtual machine.
SATySFi interpreter is slow(see #63). This bytecode compiler improves an evaluation performance.
For example, it takes about 90 minutes to make a pdf of [pi](https://twitter.com/zr_tex8r/status/973842520965656576) on my laptop by
the interpreter. By using the bytecode compiler, it only takes about 4 minutes.
The bytecode compiler can coexist with interpreter. You can use it by adding --bytecomp command line option.

To build the virtual machine, a Ruby script 'gen_code.rb' is used. This script reads 'src/frontend/bytecomp/vminstdef.yaml' and generates OCaml codes.
'vminstdef.yaml' is written in YAML format and contains instruction & primitive definitions. This mechanism makes it easier to add a new VM instruction or primtive. But now, interpreter and virtual machine are not share the code. Therefore you have to update both an interpreter and a virtual machine if you modify evaluator or add a new primitive.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,20 +4,43 @@ SRCROOT=src
 BACKEND=src/backend
 FRONTEND=src/frontend
 CHARDECODER=src/chardecoder
+BYTECOMP=$(FRONTEND)/bytecomp
 OCB_FLAGS = -cflags -unsafe-string -cflag -w -cflag -3 \
 	-use-ocamlfind -use-menhir \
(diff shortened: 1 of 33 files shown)


gfngfn:
Thank you so much for sending a really amazing PR! I confirmed that the compiler and the VM work fine. I merged the PR after making the interpreter and the VM share some duplicate code and the definition of primitive operations.

PR #82, opened 2018-06-03
 Fixed lexer not to ignore beginning-of-line spaces

matsud224:
In some cases, beginning-of-line spaces are ignored by lexer. Therefore, position information in error message may be incorrect.

For example, modify demo.saty line 70.
```
    +consol(`! Undefined control sequence.`);
```
```
  reading 'demo.saty' ...
! [Type Error] at line 70, characters 0-7:
    undefined variable '+consol'.
```
Shown error position above is not correct. It must be 'characters 4-11'.

This PR fixes this problem.

### DIFF
--- a/src/frontend/lexer.mll
+++ b/src/frontend/lexer.mll
@@ -1,6 +1,7 @@
 {
   open Types
   open Parser
+  open Lexing
 
   exception LexError of Range.t * string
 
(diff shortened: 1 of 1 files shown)


gfngfn:
Thank you for sending a PR! I have recognized the shown position is incorrect for some cases, but could not detect exactly what causes it. Let me check your fix for a while.

gfngfn:
Confirmed that it works fine. Thank you!

PR #85, opened 2018-06-07
Bit operations

konn:
This pull-request implements bit-wise operations for integers.


nekketsuuu:
Side comment: As you know, OCaml Pervasive's corresponding bitwise operations are `(lsl)`, `(lsr)`, `(lxor)`, `(land)`, `(lor)` and `lnot`. (This comment is just a note to gfngfn. I don't think we must use the same names. The naming convention of primitives is somehow bikeshed.)

gfngfn:
I’m so sorry for the extremely late response… Although the naming of the primitives may still have room for consideration, I would like to merge this PR for now, since it will be better to provide there functionalities than not, regardless of what names they have.

It may well be worth reconsidering the naming, however, when breaking changes are made in SATySFi before releasing `v0.1.z`. 

Thank you for the enhancement!

PR #92, opened 2018-06-22
Add name suggestion on error message

matsud224:
```
! [Type Error] at line 16, characters 11-19:
    undefined variable '\SATiSFy'.
    Did you mean '\SATySFi'?
```


gfngfn:
Confirmed it works. Thank you for the improvement!

PR #95, opened 2018-07-08
Fix docs regarding recent breaking changes.

qnighy:
Recent changes broke some portions in the docs. This patch is a minimal aid to fix these obvious errors. It doesn't go further e.g. adding images to let `tests/first.saty` compile.


gfngfn:
Thank you for being aware of such slight errors.

PR #96, opened 2018-07-14
Check doc builds

qnighy:
This PR is a followup for #93 and #95. Just tries to build the docs, in addition to the interpreter itself.

Let's prevent same breakages from repeatedly occurring.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -1,10 +1,22 @@
-#DEPS=pervasives.satyh gr.satyh list.satyh tabular.satyh math.satyh color.satyh deco.satyh decoset.satyh vdecoset.satyh stdja.satyh local.satyh
+#!/usr/bin/make -f
 
-all:
-	satysfi doc-primitives.saty -o primitives.pdf
+SATYSFI = satysfi
 
-lang:
-	satysfi doc-lang.saty -o lang.pdf
+all: primitives.pdf lang.pdf math1.pdf
 
-math1:
-	satysfi math1.saty -o math1.pdf
+clean:
+	$(RM) primitives.pdf lang.pdf math1.pdf $(wildcard *.satysfi-aux)
+
+primitives.pdf:
+	$(SATYSFI) doc-primitives.saty -o $@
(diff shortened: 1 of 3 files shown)


qnighy:
I started to think that it would be good to wait for #97 and then simplify this PR.

doc/Makefile
```diff
@@ -1,10 +1,22 @@
-#DEPS=pervasives.satyh gr.satyh list.satyh tabular.satyh math.satyh color.satyh deco.satyh decoset.satyh vdecoset.satyh stdja.satyh local.satyh
+#!/usr/bin/make -f
 
-all:
-	satysfi doc-primitives.saty -o primitives.pdf
+SATYSFI = satysfi
 
-lang:
-	satysfi doc-lang.saty -o lang.pdf
+all: primitives.pdf lang.pdf math1.pdf
```
omasanori: I'd like to share with you a piece of code from my WIP branch for the same goal:
```make
.PHONY: all
all: primitives.pdf lang.pdf math1.pdf

.SUFFIXES: .saty .pdf
.saty.pdf:
        satysfi $< -o $@
```

na4zagin3: FYI, suffix rules is obsolete now (v. https://www.gnu.org/software/make/manual/html_node/Suffix-Rules.html) in favor of pattern rules like this.

```make
.PHONY: all
all: primitives.pdf lang.pdf math1.pdf

%.pdf: doc-%.saty
        satysfi $< -o $@
```

omasanori: Ah, ok, I'm fine if it is also supported by non-GNU variants too. math1.saty is not doc-prefixed, so we will need renaming anyway.

na4zagin3: Oh, I didn't realized POSIX standard does not contain pattern rules, therefore NetBSD make, OpenBSD make, and so on don't support it. (FreeBSD make does and Mac OS X uses GNU make, though)

It makes sense to stick at suffix rules in terms of portability.

Issue #17, opened 2018-02-10
`cd tests ; make` fails due to missing files


na4zagin3:
### Problem
Test files under `test` requires the following files which does not exist.
- imported.pdf
- naruse1.jpg

Furthermore, it requires several fonts which are not automatically installed. (It is related to #13)
- "KozMin": KozMinPro-Regular.otf
- "Hlv": HelveticaBlack.ttf
- "ipaexm": ipaexm.ttf
- "Osaka": Osaka.ttf
- "Arno": ArnoPro-Regular.otf
- "ArnoIt": ArnoPro-Italic.otf

### Environment
- macOS 10.13.2
- OCaml 4.05.0 + OPAM 1.2.2
- SATySFi commit bc30f33

### Workaround
- Create required PDF and JPEG files under `test` directory.
- Manually copy required fonts into `lib-satysfi/dist/fonts` directory. I can copy other fonts and modify `lib-satysfi/dist/hash/fonts.satysfi-hash` to change the correspondence between internal font names and actual font paths.

omasanori:
Concerning fonts and their licensing:
- Kozuka Mincho Pro, Helvetica, Anno Pro: non-free, available from distributors (some differences between OS-bundled ones and ones from font distributors might be there)
- Osaka: non-free, bundled into macOS and predecessors

Examples of possible freely-distributable alternatives:
- Kozuka Mincho Pro: Noto Serif CJK, as a Mincho font other than IPAex Mincho
- Helvetica: TeX Gyre Heros, or other sans-serif fonts such as Liberation Sans, Roboto, Noto Sans, etc.
- Anno Pro: [EB Garamond, Junicode](https://github.com/gfngfn/SATySFi/issues/13#issuecomment-364635923), TeX Gyre Bonum, TeX Gyre Pagella, etc., as an old-style serif font
- Osaka: IPAex Gothic, M+, Noto Sans CJK, etc., as a Japanese sans-serif font

na4zagin3:
It's been fixed. Closing.

PR #97, opened 2018-07-14
Download Junicode and IPAex fonts by `make lib`.

omasanori:
Both font families are distributed under OSI-approved licenses and we have no restriction against automatic installation.

I added `-L` option to curl to follow redirection. It actually needed only for Junicode, but it is added to all curl commands for consistency.

Related to: https://github.com/gfngfn/SATySFi/issues/17 https://github.com/gfngfn/SATySFi/issues/38


omasanori:
This PR should be revised after https://github.com/gfngfn/SATySFi/pull/108 is merged.

PR #99, opened 2018-07-14
rewrite gen_code.rb in OCaml

leque:
I rewrote `gen_code.rb` in OCaml. This is an almost literal porting. Generated code is identical to before.

pros:

* build without ruby
* VM instruction definitions are statically typed

cons:

* `vminst.ml` are not quite accessible from other languages (but who cares? :) )
* will break some PRs

You can check generated code as below:

```
# generate files with the ruby version in the master branch
git checkout master
make gen

# checkout this PR
git checkout gencode_ml

# commit the ruby version output for diff
git add -f src/frontend/*.gen.ml src/frontend/bytecomp/*.gen.ml
git commit -m tmp

# re-generate files with the ocaml version
make clean gen

# check diff (will print nothing)
git diff
```


leque:
rebased to the current master.

leque:
OK, I have made this PR more conservative for ease of review and rebasing. `dune` migration and removing ruby related files will be an other PR.

Makefile
```diff
@@ -27,35 +27,43 @@ GENS= \
...
 
+$(INSTDEF): $(INSTDEF_YAML)
+	$(RUBY) $(GENCODE_RB) --ml $< > $@
+
 gen: $(GENS)
 
-$(ATTYPE_GEN): $(INSTDEF) $(GENCODE)
-	$(RUBY) $(GENCODE) --gen-attype $(INSTDEF) > $@
+$(ATTYPE_GEN): $(INSTDEF)
+	$(GENCODE) --gen-attype > $@
```
nekketsuuu: How about generating codes by dune instead of make? For example, dune generates `__attype.gen.ml` if adding the following lines to `src/dune`.

```
(rule
 (targets __attype.gen.ml)
 (deps    (:gen ../tools/gencode/gencode.exe))
 (action  (with-stdout-to %{targets}
           (run %{gen} --gen-attype))))
```

(Ref: [Using a custom code generator](https://jbuilder.readthedocs.io/en/latest/quick-start.html?highlight=preprocess#using-a-custom-code-generator) -- dune’s documentation)

KeenS: this?

>  dune migration and removing ruby related files will be an other PR.

nekketsuuu: Oh I'm sorry I skipped the mention of `dune`.

leque:
Thanks. I have created #142.

PR #100, opened 2018-07-14
Fix types of documented primitives

na4zagin3:
This PR is to fix types in `doc/doc-primitives.saty` because of `(-->)` being left-associative.


gfngfn:
Thank you!

PR #102, opened 2018-07-16
setup ocp-indent

leque:
resolve #39.

I have no particular preference on indentation, so I choose the same style as [one used in the OCaml compiler](https://github.com/ocaml/ocaml/blob/47eeff4bac81e710489147bbff41d23d38d4d5c2/.ocp-indent).

About ocp-indent, see [its documentation](https://www.typerex.org/ocp-indent.html).

PR #103, opened 2018-07-16
Check coding style using ocp-indent with Travis CI.

omasanori:
To keep consistent coding style, it is nice to check it automatically, always.

Depends on: https://github.com/gfngfn/SATySFi/pull/102

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,12 @@ install:
   - wget https://raw.githubusercontent.com/ocaml/ocaml-travisci-skeleton/master/.travis-ocaml.sh
   - bash -ex .travis-ocaml.sh
   - eval `opam config env`
+  - opam install -y ocp-indent
   - opam repository add satysfi-external https://github.com/gfngfn/satysfi-external-repo.git
   - opam update
   - opam pin add -y --no-action satysfi .
   - opam install -y --deps-only satysfi
 script:
+  - find src -name '*.ml' -o -name '*.mli' | xargs ocp-indent -i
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -9,9 +9,11 @@ install:
   - wget https://raw.githubusercontent.com/ocaml/ocaml-travisci-skeleton/master/.travis-ocaml.sh
   - bash -ex .travis-ocaml.sh
   - eval `opam config env`
+  - opam install -y ocp-indent
   - opam repository add satysfi-external https://github.com/gfngfn/satysfi-external-repo.git
   - opam update
   - opam pin add -y --no-action satysfi .
   - opam install -y --deps-only satysfi
 script:
+  - diff -u <(printf '') <(find src/ -name '*.ml' -or -name '*.mli' | xargs -I{} sh -c 'ocp-indent {} | diff -u {} -')
```
leque: or, more simply
```
- find src -name '*.ml' -o -name '*.mli' | xargs ocp-indent -i
- git diff --quiet
```

omasanori: I wanted to avoid writing files for performance gain, but I will measure it later.

omasanori: @leque Your method is actually faster than mine. I tested on WSL, so it may change on Linux on container but anyway it seems worth little. I will revise it. Thanks!

omasanori:
Closed as it is too old. I hope a kind of automatic formatting enforcement will be land, though.

Issue #21, opened 2018-02-11
Switch to dune


objmagic:

qnighy:
In #54 I'm trying to introduce custom build rules using `myocamlbuild.ml`. In order to have both #21 and #54, we will have to write the same rules for dune.

PR #105, opened 2018-08-24
Fix module names with underscore suffix

matsud224:
Because of #21, we can now remove underscore suffixes in some module names.


gfngfn:
Merged. Thank you!

Issue #106, opened 2018-08-29
Sandboxing of OPAM 2 prevents the build from downloading font files


camlspotter:
OPAM2 now has sandboxing: the build cannot access the external resources.  SATySFi build fails at the downloading of font files.

See https://github.com/ocaml/opam/issues/3460 for a discussion of the sandboxing.

In the following build attempt, `curl` fails due to the name resolution failure of www.gust.org.pl.
```
$ opam --version
2.0.0
$ opam switch show
/Users/jun/.share/4.06.1-satysfi
$ opam pin add .
[NOTE] Package satysfi is currently pinned to
       git+file:///Users/jun/.share/4.06.1-satysfi/SATySFi#master (version
       0.0.2).
[satysfi.0.0.2: git]
[satysfi.0.0.2: git]
[satysfi.0.0.2: git]
[satysfi.0.0.2: git]
[satysfi.0.0.2: git]
[satysfi.0.0.2: git]
[satysfi.0.0.2] synchronised from git+file:///Users/jun/.share/4.06.1-satysfi/a/SATySFi#master
satysfi is now pinned to git+file:///Users/jun/.share/4.06.1-satysfi/a/SATySFi#master (version 0.0.2)
The following actions will be performed:
  ∗ install satysfi 0.0.2*
Do you want to continue? [Y/n] 
y

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫 
[ERROR] The compilation of satysfi failed at
        "/Users/jun/.opam/opam-init/hooks/sandbox.sh build make -f Makefile lib
        PREFIX=/Users/jun/.share/4.06.1-satysfi/_opam".

#=== ERROR while compiling satysfi.0.0.2 ======================================#
# context              2.0.0 | macos/x86_64 | ocaml-base-compiler.4.06.1 | pinned(git+file:///Users/jun/.share/4.06.1-satysfi/a/SATySFi#master#285ff60f)
# path                 ~/.share/4.06.1-satysfi/_opam/.opam-switch/build/satysfi.0.0.2
# command              ~/.opam/opam-init/hooks/sandbox.sh build make -f Makefile lib PREFIX=/Users/jun/.share/4.06.1-satysfi/_opam
# exit-code            2
# env-file             ~/.opam/log/satysfi-87558-335865.env
# output-file          ~/.opam/log/satysfi-87558-335865.out
### output ###
# if [ -x "$(command -v curl)" ]; then \
# [...]
[...]

PR #108, opened 2018-09-18
Migrate to opam 2.0.

omasanori:
[opam 2.0.0 was released](http://opam.ocaml.org/2.0-preview/blog/opam-2-0-0/)! Now it is time to go forward.

Fixes https://github.com/gfngfn/SATySFi/issues/106

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,20 +71,6 @@ install: $(TARGET)
 #preliminary:
 #	[ -d .git ] && git submodule update -i || echo "Skip git submodule update -i"
 
-lib:
-# -- downloads fonts --
-	mkdir -p temp/
-	if [ -x "$$(command -v curl)" ]; then \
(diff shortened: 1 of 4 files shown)


omasanori:
Oops, I forgot to update README too. Please do not merge until it is fixed.

omasanori:
Note that Homebrew opam formula has not upgraded to 2.0 yet. (Official installation script works as Linux.)

nekketsuuu:
Sounds great! It seems that OPAM2 uses `eval $(opam env)` instead of ``eval `opam config env` `` (See [FAQ](http://opam.ocaml.org/2.0-preview/doc/FAQ.html)). So we should fix instructions in README.md and README-ja.md.

omasanori:
Good catch! I will fix it.

PR #109, opened 2018-09-19
Use $BINDIR/../share/satysfi.

omasanori:
I confirmed that `rm -rf $HOME/.satysfi; cd doc; make` succeeded with this change + https://github.com/gfngfn/SATySFi/pull/108 using opam 2.0.0.

Fixes https://github.com/gfngfn/SATySFi/issues/38

### DIFF
--- a/src/frontend/main.ml
+++ b/src/frontend/main.ml
@@ -745,7 +745,7 @@ let setup_root_dirs () =
       | None    -> []
       | Some(s) -> [s]
     else
-      ["/usr/local/share/satysfi"; "/usr/share/satysfi"]
+      [Filename.concat (Filename.dirname Sys.executable_name) "../share/satysfi"; "/usr/local/share/satysfi"; "/usr/share/satysfi"]
   in
   let user_dirs =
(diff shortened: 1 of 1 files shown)


omasanori:
@gfngfn @na4zagin3 Does this PR become irrelevant with recent development by you regarding packaging and deployment?
If so, I'm happy to close this and make the number of open PRs fewer. 😃

If it is still relevant, could you give me a suggestion to make it integrated into master?

na4zagin3:
IMO, it is not a problem now as there will be only two kinds of novice users, I think.
1. Users following the official readme will have the library under `/usr/local/share/satysfi`
2. Users using Satyrographos will have the library in the opam registory

I would like to hear @gfngfn’s thought.

na4zagin3:
This PR should be irrelevant now because `satysfi.opam` does not install dist files.

FYI, discussion on search paths is also ongoing in https://github.com/gfngfn/SATySFi/pull/212.

omasanori:
I see. I am glad to hear that the problem finally went away. 😃

PR #114, opened 2018-10-19
math.satyh: change \centerdot to \cdot

nekketsuuu:
This PR changes the name of `\centerdot` to `\cdot`, which is the same name with TeX's amsmath.

I don't know why, but `\centerdot` from TeX's [amssymb](https://ctan.org/pkg/amsfonts) is equivalent to a square dot `\sqbullet`, not a circular dot `\cdot`. Also, TeX's [unicode-math](https://ctan.org/pkg/unicode-math) doesn't provide `\centerdot`.


gfngfn:
I understand that. Thanks for modification!

PR #116, opened 2018-10-19
README.md: add instructions to install OPAM without bwrap

### LINKED ISSUE #115: Add instruction for WSL users
As reported in https://github.com/ocaml/opam-repository/issues/12050#issuecomment-393478072, we have to disable bubblewrap to use OPAM on WSL. I believe that problem confuses many newbies who try to install SATySFi on WSL. (e.g., https://twitter.com/puripuri2100/status/1053254602273959936)

It’d better to have a notice or an instruction for WSL users.

nekketsuuu:
fix #115


gfngfn:
Thank you!

PR #123, opened 2018-11-04
Support fixed-length list pattern

nekketsuuu:
This PR implements fixed-length list pattern.

Example:

```satysfi
@require: standalone
@require: stdja

let show lst =
  let-rec aux
    | [] = ` `
    | [x] = x  % <-- This line!
    | (x :: xs) = x ^ `; `# ^ (aux xs)
  in `[` ^ (aux lst) ^ `]`
let-inline \show-string str = embed-string str

in
standalone '<
  +p {
    \show-string(show [`a`; `b`; `c`]);
  }
>
```

Without this, we must write a fixed-length list explicitly by cons and nil (e.g. `x :: []` instead of `[x]`). In OCaml, we can use this kind of syntactic sugar.


gfngfn:
Thanks for the improvement!

PR #126, opened 2018-11-27
Add some font functions and add bold type

puripuri2100:
This PR adds font functions to some class files.

Since font's function was not provided in `std.satyh`, I will add it.

- `font-latin-roman`
- `font-latin-sans`
- `font-latin-mono`
- `font-cjk-mincho`
- `font-cjk-gothic`

Since bold type was not provided, I will add it.

- `font-latin-bold`


gfngfn:
I’m so sorry for the extremely late response… Thank you for the improvement!

PR #130, opened 2018-12-06
Fix the registration of section labels in stdjareport

elpinal:
The current implementation of `+section` in `stdjareport` is a bit wrong so that `\ref` does not work well for section labels. This PR fixes the problem.


gfngfn:
Thank you for fixing!

PR #133, opened 2018-12-10
Ordering relation on strings

elpinal:
Ordering relations on strings are needed to define sort functions on strings.
This PR adds 1 primitive and 1 module.

### Added primitive

- `string-lt : string -> string -> bool`
This primitive tests whether the first argument is less than the second argument.

### Added module

- `String`
This module provides operations on strings:
  - `equal`
equal
  - `lt`
less than
  - `gt`
greater than
  - `le`
less than or equal
  - `ge`
greater than or equal

### Note

It would be arguable whether this design is good or not.

### DIFF
--- a/lib-satysfi/dist/packages/string.satyg
+++ b/lib-satysfi/dist/packages/string.satyg
@@ -1,0 +1,22 @@
+
+module String
+: sig
+    val equal : string -> string -> bool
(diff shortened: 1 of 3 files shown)


konn:
Just a margial note: seems related to #60.

PR #135, opened 2018-12-12
fix \ref in stdjareport

matsud224:
`\ref` in `stdjareport` was fixed in #130 but it seems to be still wrong.
* When `section-scheme` registers section number or page number with `register-cross-reference`,  extra string `section:` is added in front of label. `\ref` command doesn't add `section:` to label when search label with `get-cross-reference`. So, it couldn't find appropriate section number or page number.
* `Annot.register-location-frame` isn't called in `chapter-scheme`. So, hyperlink doesn't work. (I missed `chapter-scheme`. Sorry!)


gfngfn:
Merged. Thank you!

PR #137, opened 2019-01-12
Improve installation guide

azwjp:
When installation, SATySFi requires to move some files and download fonts, but there's no explanation in README.
Additionally, demo file is not able to be compiled because of lacks a image file.


nekketsuuu:
Related: #38, #72, and #109.

gfngfn:
I replaced the description of manual installation of fonts with that of invoking shell scripts `download-fonts.sh` and `install-libs.sh`. Thank you anyway for the improvement!

PR #139, opened 2019-01-13
[RFC] Support processing Omd.Img for Markdown image

hikalium:
[This document](https://github.com/gfngfn/SATySFi/wiki/SATySFiによるMarkdown文書からのPDF出力) written by @gfngfn says that the current SATySFi does not support Markdown image intentionally.
However, implementing this functionality is quite simple as shown in this p-r. (This patch is not completed but you can see that the change is very small.) 
I want to discuss about why this functionality is not supported for now and what is the potential solution.

### DIFF
--- a/lib-satysfi/dist/md/mdja.satysfi-md
+++ b/lib-satysfi/dist/md/mdja.satysfi-md
@@ -29,6 +29,7 @@
   "code-default": "\\code",
   "url": "\\link",
   "reference": "\\reference",
+  "img": "\\img",
   "embed-block": "\\embed-block",
   "err-inline": "\\error"
 }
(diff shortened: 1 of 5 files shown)


gfngfn:
I’m sorry for the late response. The reason why I have somewhat hesitated to support images in Markdown source is that the current spec of SATySFi does not handle relative paths in a natural manner (this issue is discussed in the 8th item of [this article](https://github.com/gfngfn/SATySFi/wiki/SATySFi-v0.1.x%E3%81%A7%E8%A1%8C%E3%81%AA%E3%81%86%E7%A0%B4%E5%A3%8A%E7%9A%84%E5%A4%89%E6%9B%B4%E3%81%AE%E6%A1%88)).

However, I also consider that supporting the insertion of images in a slightly unnatural manner is much better than not supporting it, and now I am willing to merge this PR. Thank you for the improvement.

hikalium:
Thank you for merging this p-r and telling me the reason why Markdown image support have not been introduced.
Resolving file-path related things are so hard... I hope that issue is fixed in a user-friendly way.
Anyway, though the Markdown image support is not perfect for now, This functionality and your product are useful for many people. Thank you again for your work.

PR #142, opened 2019-02-02
remove dependency on ruby

leque:
This is follow-ups for #99.

* `*.gen.ml` files generation rules move from `Makefile` to `dune`
* remove ruby related files and description.


gfngfn:
I’m so sorry for the extremely late response… Thank you for improving the build process!

I appended some commits to this PR in order to fix conflicts between `master` (and finally the build passed):

https://github.com/gfngfn/SATySFi/compare/master..48914724140d3127a53f0f96e097112f775d9a2f

Should the modification above has any problem, I would appreciate it if you could tell me about that.

gfngfn:
Merged this PR. Thank you so much!

PR #145, opened 2019-02-06
Remove Action module

matsud224:
Camlpdf v.2.2.2+satysfi provides `Pdfaction` module. So, we can replace `Action` module in SATySFi with `Pdfaction`.


gfngfn:
Thank you. I also modified `Action` in `vminstdef.yaml` into `Pdfaction` just to be sure.

PR #147, opened 2019-02-08
OPAM: add synopsis and description

nekketsuuu:
Currently, OPAM2 produces the following warning after running `opam pin add satysfi .`:

```
[WARNING] Failed checks on satysfi package definition from source at git+file:///mnt/c/Users/nek/go/src/github.com/gfngfn/satysfi#master:
    error 57: Synopsis and description must not be both empty
```

This PR resolves this warning by adding synopsis and description copied from README.md.

See [this document](https://opam.ocaml.org/doc/Packaging.html) or [other packages' OPAM files](https://github.com/ocaml/opam-repository/tree/master/packages) for detailed explanation of these fields.

### DIFF
--- a/satysfi.opam
+++ b/satysfi.opam
@@ -72,3 +72,6 @@ depends: [
   "yojson" {= "1.4.1+satysfi"}
   "omd"
 ]
+synopsis: "A statically-typed, functional typesetting system"
+description: """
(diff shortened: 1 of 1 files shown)


gfngfn:
Thank you!

PR #161, opened 2019-02-17
add -C command line option (fix #154)

### LINKED ISSUE #154: Per-project font configuration
Currently there is no means to distribute .saty files together with custom font settings (fonts.satysfi-hash).
It is cumbersome to distribute a shell script that modifies say `~/.satysfi/local/hash` whenever I share .saty files portably.
It is helpful if satysfi uses custom hash/fonts when I put them in a git repository.
I think one way to do so is to add the current directory to satysfi's search path.
That is, satysfi will search for configs "./", "~/.satysfi", "/usr/local/share/satysfi", "/usr/share/satysfi" in this order.
Another possible idea is add a command line option like `-C path`, analogous to `make -C path` or `gcc -I path`.
My preference is the latter, because it will be much easier to write a portable Makefile that executes satysfi.

nyuichi:
Satysfi's config search path is modified as follows:

- without -C option: $(PWD)/.satysfi, $HOME/.satysfi, /usr/local/share/satysfi, /usr/share/satysfi
- with -C path1[:path2,...]: path1, [path2, ..., ] $HOME/.satysfi, /usr/local/share/satysfi, /usr/share/satysfi

### DIFF
--- a/src/backend/optionState.ml
+++ b/src/backend/optionState.ml
@@ -15,6 +15,7 @@ type state = {
   mutable debug_show_bbox : bool;
   mutable debug_show_space : bool;
   mutable mode             : (string list) option;
+  mutable extra_config_paths : string list option;
 }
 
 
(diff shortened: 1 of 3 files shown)


gfngfn:

PR #163, opened 2019-02-21
More math char classes (fix #159)

### LINKED ISSUE #159: More math character classes
Currently satysfi only supports a subset of font styles defined in unicode to be used in math expressions [2]. In [1] many (maybe all?) font styles available in unicode are listed up. Among them the typewriter style should be useful in particular.

- [1] http://qaz.wtf/u/convert.cgi?text=The+secret+is+out.+
- [2] The SatysfiBook Sect. 10.4.

nyuichi:


nyuichi:
rebased

Issue #125, opened 2018-11-05
How about using annotated tags for releases?


nekketsuuu:
Currently, git tags `v0.0.1`, `v0.0.2`, and `v0.0.3` are all lightweight tags. But it's better to use annotated tags for releases.

@gfngfn Would you use annotated tags for future releases?

* Ref.
    * [How can I tell if a given git tag is annotated or lightweight?](https://stackoverflow.com/q/40479712/5989200) -- Stack Overflow
    * [In what circumstances should I add the -a flag to the git tag command?](https://stackoverflow.com/q/4092640/5989200) -- Stack Overflow
    * [How to annotate an existing tag in git](https://stackoverflow.com/q/42390650/5989200) -- Stack Overflow
    * [`dune subst` and git lightweight tags](https://discuss.ocaml.org/t/dune-subst-and-git-lightweight-tags/2830) -- OCaml Discuss

nekketsuuu:
`v0.0.4` was released, but this is also lightweight (´・ω・)

```
$ git for-each-ref refs/tags
bb3a06c949e1d2df61571dee4ad2cc6a3aa3e98a commit refs/tags/v0.0.1
51687b6afff93257c3e24263cacc80ca34995299 commit refs/tags/v0.0.2
4e5eb43e629ff013672604cc6b010bcc2e0efe19 commit refs/tags/v0.0.3
68b2b150e7caf33ab52de58327ee3662e6bc2e43 commit refs/tags/v0.0.4
d0e91b0986b655cd6a1933dad56dd9b8661dadb3 tag    refs/tags/vm0.9
a0459c65c54a907b7979f1b78bbd26cd4d4087b5 tag    refs/tags/vm1.0b
08cba2265a06896119932a10f8befe7984a7f23d tag    refs/tags/vm1.0g
e7a37b962c949ef64f421f7b9d0312a32324b7a2 tag    refs/tags/vm1.0z
```

gfngfn:
Ah, I had forgotten this issue! 🙏

nekketsuuu:
`v0.0.5` was released using an annotated tag. Thanks gfn-san!

PR #167, opened 2019-02-28
[WIP] Version watermarking by Dune

nekketsuuu:
## What's new?

This PR implements more detailed version strings.

```sh-session
$ opam pin remove satysfi
$ git clone -b nekketsuuu-watermarking https://github.com/nekketsuuu/SATySFi.git
$ cd SATySFi
$ git tag -a -m "virtual release for testing" v0.0.4
$ opam pin add satysfi .
(omit.)
$ satysfi --version
  SATySFi version v0.0.4
$ touch foobar
$ git add foobar
$ git commit -m "test commit"
[nekketsuuu-watermarking b0b3c65] test commit
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 foobar
$ opam reinstall satysfi
(omit.)
$ satysfi --version
  SATySFi version v0.0.4-1-gb0b3c65
$ opam pin remove satysfi  # Confirm to remove the tag "v0.0.4" from an OPAM cache
```

In the version string `v0.0.4-1-gb0b3c65` in the nearly last line, `1` means that it is 1 commit ahead from v0.0.4, and `gb0b3c65` means its commit hash "b0b3c65..." by **g**it. This is the same as the output of `git describe --dirty --always`.

By this PR, we can:

* easily show an accurate version string, and
* easily update a version string (it's unnecessary to update the OPAM file).

### Miscellaneous notes

* Currently, `satysfi --version` outputs `SATySFi version vm1.0z-****-g****` because tags `v0.0.1`, `v0.0.2`, and `v0.0.3` are lightweight tags (#125). `vm1.0z` is an annotated tag for old Macrodown. So this is not a bug of this PR.
    * This is why I made this PR as a draft pull request.
* This PR uses `dune subst`, which is documented [here](https://dune.readthedocs.io/en/latest/usage.html#watermarking).
* The version string is properly replaced while running `opam pin add`, `opam reinstall`, or `dune-release`. In contrast, simple `make` or `dune build` doesn't replace it, and show `%%VERSION%%` for the version string. This is intended; I think the version string is confusing and not needed for the development period. If needed, use `opam pin add` or `opam reinstall`.
* It is still under discussion how to handle versions by Dune (ocaml/dune#880). Although I believe my method is better for SATySFi to handle versions, it will be worth reading that issue for future possible changes of Dune.


nekketsuuu:
https://github.com/gfngfn/SATySFi/issues/125 is resolved and this pull req is now ready for review :)

satysfi.opam
```diff
@@ -10,6 +9,7 @@ dev-repo: "git+https://github.com/gfngfn/SATySFi.git"
 bug-reports: "https://github.com/gfngfn/SATySFi/issues"
 build: [
   ["mkdir" "-p" "temp"]
+  ["dune" "subst"] {pinned}
```
na4zagin3, line 12: I'm curious. Why is this line executed only during pinning, but not installing non-pinned versions?

na4zagin3: I confirmed `%%VERSION%%` was not substituted when it was installed as a non-pinned package.

```
# Adding a package pointing to `git+https://github.com/nekketsuuu/SATySFi.git#68f8de124503634b05608f993b693dcf792d437b` to a local repository
$ opam install satysfi
$ satysfi --version
  SATySFi version %%VERSION%%
```

nekketsuuu: Please read https://github.com/gfngfn/SATySFi/pull/167#issue-257002956

> This is intended; I think the version string is confusing and not needed for the development period. If needed, use `opam pin add`.

na4zagin3: Oops, Ioverlooked the paragraph. Sorry.  So, this PR changes the release process as well: `dune-release tag` instead of `git tag -a`. It will be helpful if the new release process is explicitly described somewhere in the repo.  Just my two cents.

nekketsuuu: No, this pull request doesn't change the release process of SATySFi. Just running `git tag -a` is enough.

nekketsuuu: This is because SATySFi is now installed/built using opam pin: https://github.com/gfngfn/SATySFi/blob/2271124d38ebd53d942e0cd05a3edfa958928e15/README.md#build

na4zagin3: Well, there are non-negligible number of users explicitly and implicitly installing SATySFi from an opam repo, namely [satyrographos-repo](https://github.com/na4zagin3/satyrographos-repo), including https://hub.docker.com/r/amutake/satysfi (pulled 7.8k times).  Although I can patch this line, I'd like not to add inessential patches.

BTW, I was wrong about `dune-release tag` which in fact does not replace the string.  It's required to run `dune-release` to upload a tarball as release asset.

Didn't you recommended @gfngfn adds SATySFi to the official OPAM repo?  I'd leave a notice somewhere so that he doesn't fall into the trap then.  Should I file an issue instead of mentioning it in the code base?

nekketsuuu: > Didn't you recommended @gfngfn adds SATySFi to the official OPAM repo?

This is out of scope of this pull req. This pull request just makes it easy to confirm which version of SATySFi is used if HEAD is not tagged (i.e. not released) by adding its commit hash to the version string.

nekketsuuu: @na4zagin3 Is there any proposed changes from you? Note that if I remove `{pinned}`, then `%%VERSION%%` is replaced every time after building SATySFi for development, and this process creates an unnecessary diff which should not be staged.

na4zagin3: It sounds totally reasonable to me to leave the line as is (i.e., with `{pinned}`); I'd like to have notes on the release process _somewhere,_ for example, a section in the README, or a new file:

> ### Release Process
> SATySFi uses dune-subst to substitute watermarks. Therefore it must be released in the following process so that prepared distributions (e.g., `satysfi-v0.0.6.tbz`) are uploaded to GitHub.
> 
> ```
> $ git tag add -v <version>
> $ dune-release publish distrib # or something like that
> ```

nekketsuuu: Then it's out of scope of this pull request. I don't intend to change the way to release a new version. `dune-release` is not fit to current SATySFi.

na4zagin3: It _is_ a consequence of this PR, which changes how SATySFi should be released.

As OPAM repos prefers a source tarball in `url` section rather than source repos (especially, the official OPAM repo _prohibits_ source repos there), there are only two ways to substitute watermarks:

1. Release a prepared distribution, in other words, source code preprocessed by dune-subst. 
2. Upload a OPAM file which has a line like `["find" "." "-iname" "*.ml" "-exec" "sed" "-i.bak" "-e" "s/%%VERSION%%/v0.0.6/" "{}" ";"]` instead of `["dune" "subst"] {pinned}`.

Do you think “`dune-release` is not fit to current SATySFi” because it automatically attempts to release SATySFi to GitHub and to send a PR to the official OPAM repository?

I edited my previous comment with `dune-release distrib` that only generates a prepared distribution.  Does that make sense to you?

na4zagin3: Hmm, even though it is a consequence, it can be handled by another issue/PR.  You're right.  Shall I file one?

na4zagin3: I filed https://github.com/gfngfn/SATySFi/issues/237 since I don't want to block you on this. Today I learned a lot.  Thank you for a fruitful discussion!

nekketsuuu: I don't think this pull req changes the current release process, though.... I commented my opinion to that issue <https://github.com/gfngfn/SATySFi/issues/237#issuecomment-660766570>. Let's continue this discussion there.

nekketsuuu:
It turns out by @na4zagin3 that this pull req removes a concrete version string from assets for a git tag, which is currently automatically created by GitHub (e.g. <https://github.com/gfngfn/SATySFi/releases/tag/v0.0.5>).

There are three ways to handle this issue.

* Don't worry about this change. It's completely OK if `%%VERSION%%` is remained in the assets. SATySFi itself never uses them.
* Change the release process of SATySFi. For example, use dune-release to replace `%%VERSION%%` in the assets.
* Close this pull req without merging.

@gfngfn Which resolution do you prefer? Any method is fine with me. Since you are responsible for releasing new versions of SATySFi, you can choose how to handle this issue =)

na4zagin3:
@gfngfn FYI, if you chose the second option, there would be two methods:

### with `dune-release distrib`

After pushing an annotated-tag to GitHub as usual, run `dune-release distrib` and  [manually register](https://docs.github.com/en/github/administering-a-repository/managing-releases-in-a-repository#editing-a-release) the prepared tarball.  I wrote this down at https://github.com/gfngfn/SATySFi/issues/237

### with `dune-release`

After adding an annotated-tag, run `dune-release` to upload the tag and the prepared tarball to SATySFi repo in GitHub, then abort registering SATySFi to the officlan OPAM repo.

na4zagin3:
BTW, why not use `%%VERSION_NUM%%` rather than `%%VERSION%%`?  Is it intended to report the version with `v` prefix (e.g., `v0.0.5`)?

PR #168, opened 2019-03-05
support subset embedding of OpenType CFF fonts

matsud224:
This PR aims to follow otfm changes of gfngfn/otfm#6.
When we use an OpenType CFF font, SATySFi embeds whole font file. This causes expansion of PDF file size. This PR and gfngfn/otfm#6 solve this problem by embedding only necessary glyphs and related data.


gfngfn:
Merged this PR. Thank you very much!

PR #171, opened 2019-03-06
fix a compare function of Set

matsud224:
This PR replaces a compare function of ElementSet (in exhchecker.ml) with Pervasives.compare. Previous definition of compare was not a total order.


gfngfn:
I’m super sorry for the super late response… I understand the modification. Thank you for fixing it!

PR #173, opened 2019-03-12
build src/* as a private library

leque:
This PR builds files under `src/` as a private package.

This make it easy to write OCaml level test cases for SATySFi. I wrote some parser tests and refactoring as a proof of concept of this patch.


gfngfn:
I’m super sorry for the extremely late reply… The modification looks brilliant as well.

I merged this PR to a new branch stemming from the current `master` and resolved small conflicts (https://github.com/gfngfn/SATySFi/commit/6c84affd8cd47fd9ada6188dd2dd920469aa3ff3), and tried to build the executable and to run tests. The build by just invoking `make` was successful, but tests did not seem to be run by `dune runtest` nor by `dune build test/parser_test.exe && dune runtest` (the commands emit nothing on stdout).

I consulted [the documentation of Dune](https://dune.readthedocs.io/en/stable/tests.html), but I haven’t figure out how to run tests in this case yet. Would you mind telling me how the tests are intended to be run?

Et7f3:
Consider adding empty mli for main binary as it reduce size. Dune does this https://github.com/ocaml/dune/blob/b2bfb9829b0c7d39a64d0467c0989b7885100831/bin/main.mli#L1

leque:
> Consider adding empty mli for main binary as it reduce size.

Thank you for pointing out this. It could be automatically done by dune. https://github.com/ocaml/dune/pull/3768 (We need to update `dune-project` to `(lang dune 2.8)` to ensure this behavior [including dune 2.x](https://github.com/ocaml/dune/pull/3768#issuecomment-692866162).)

Et7f3:
I think in recent version of dune public binary can depend on private library (I see one lib is exposed I don't know if it is intended to be)

PR #176, opened 2019-03-17
Report ParseErrorDetail in common format

hanazuki:
For ease of editor support, this patch changes the reporting format for `ParseErrorDetail` so that the range of error is printed on the same line with the "!" marker, as for other types of errors.

I'm developing Emacs support for SATySFi and find it would be helpful if all kinds of errors would be reported in the same format to highlight error locations correctly in the editor.

### DIFF
--- a/src/frontend/main.ml
+++ b/src/frontend/main.ml
@@ -608,7 +608,11 @@ let error_log_environment suspended =
       ]
 
   | Parsing.Parse_error -> report_error Parser [ NormalLine("something is wrong."); ]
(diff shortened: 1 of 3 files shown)


gfngfn:
I am so sorry for the late response, and thank you for looking into the implementation.

I understand what the modification is for, and confirmed that the modified implementation works fine. (The function `report_error`, however, is no longer used in the parser, so it can be entirely removed.)

Thank you for improvement!

PR #180, opened 2019-05-24
Fixes for netprint

matsud224:
[netprint](https://www.printing.ne.jp/) is a printing service in Japan. We can upload files via internet and print them at seven-eleven (convenience store).
netprint doesn't accept PDFs generated by SATySFi and shows the following error:
> 登録エラーが発生しました。 サポートしているファイル名・ファイル形式をご確認ください。(1699)  

This PR fixes this problem by suppressing null-value entries of /DW & /DW2 in Font Dictionary. /DW & /DW2 are optional entries and have integer and array type respectively. PDF 1.7 spec section 3.2.8 says:
> Specifying the null object as the value of a
dictionary entry (Section 3.2.6, “Dictionary Objects”) is equivalent to omitting
the entry entirely.

Null-value entries of /DW & /DW2 may not break the specification but netprint is sensitive about it.

### DIFF
--- a/src/backend/fontFormat.ml
+++ b/src/backend/fontFormat.ml
@@ -1419,6 +1419,13 @@ let of_per_mille_pair_opt = function
   | Some((PerMille(a), PerMille(b))) -> Pdf.Array[Pdf.Integer(a); Pdf.Integer(b)]
 
 
+let add_entry_if_non_null key value dict =
+  if value <> Pdf.Null then
+    (key, value) :: dict
+  else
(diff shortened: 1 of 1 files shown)


gfngfn:
I’m so sorry for the late response…

I understand the purpose for which this PR was made, read the diff, and confirmed that the modified program works well.

I’m slightly curious about how you detected that the cause of errors emit by netprint was null objects, but anyway, thank you for improvement!

PR #189, opened 2019-09-08
Fix an error message

elpinal:
A space is missing in an error message.


gfngfn:
I’m sorry for the late response. Thanks for fixing!

PR #190, opened 2019-09-08
Add floating-point operations, ceil and floor

elpinal:
This PR adds 2 floating-point operations:

- `ceil : float -> float`
- `floor : float -> float`

### DIFF
--- a/tools/gencode/vminst.ml
+++ b/tools/gencode/vminst.ml
@@ -2674,6 +2674,36 @@ make_float (atan flt1)
         ~code:{|
 make_float (atan2 flt1 flt2)
 |}
+    ; inst "PrimitiveCeil"
+        ~name:"ceil"
+        ~type_:{|
+~% (tFL @-> tFL)
+|}
+        ~fields:[
+        ]
+        ~params:[
+          param "fc1" ~type_:"float";
+        ]
+        ~is_pdf_mode_primitive:true
+        ~is_text_mode_primitive:true
+        ~code:{|
+make_float (ceil fc1)
+|}
+    ; inst "PrimitiveFloor"
+        ~name:"floor"
+        ~type_:{|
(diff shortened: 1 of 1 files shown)


nyuichi:
@elpinal 

Nice pull request 👍 
Could you possibly add some more functions like the following too?

- `truncate : float -> float`
- `round : float -> float`
- `pow : float -> float -> float`      (pow x y = x^y)
- `exp : float -> float`        (exp x = e^x)
- `sqrt : float -> float`
- `log : float -> float -> float`
- `log10 : float -> float`

elpinal:
@nyuichi SATySFi already has a primitive named `round`, with type `float -> int`. SATySFi's `round` is defined as OCaml's `int_of_float`, which in turn is equal to OCaml's `truncate`. What do you think of that?

nyuichi:
@elpinal 

Thanks for your feedback!

> SATySFi already has a primitive named round, with type float -> int.

Aw maybe we need to use a new name other than round. One better name is `float-round`, I think. (cc: @gfngfn)

> SATySFi's round is defined as OCaml's int_of_float, which in turn is equal to OCaml's truncate

What I want is the function called `Float.trunc` in OCaml's standard library, which is not equivalent to `truncate` in Pervasives.
I would name this `float-truncate` following the above convention.
To be honest, their actual names do not matter to me because I will just re-export them in `satysfi-li` as `Float.round` and `Float.truncate`.
So I don't have a strong opinion about this.

gfngfn:
I’m so sorry for the late response… I merged only the implementation of `ceil` and `floor` for now.

Note: currently, `tools/gencode/vminst.ml` is automatically generated from `src/frontend/bytecomp/vminstdef.yaml`. Thus I updated the latter so that it generates the former one, which you had extended. We will probably put only the former under version control in the future.

PR #192, opened 2019-09-14
math.satyh: '\approx' is defined twice

elpinal:
In `math.satyh`, there are 2 same specifications and 2 same definitions for `\approx`. This PR removes the duplicates.


gfngfn:
I am so sorry for the late reply. Thanks!

PR #194, opened 2019-09-15
Introducing semantic signatures

### LINKED ISSUE #193: Arity mismatch of type constructors should be detected as soon as possible
As written in a comment in `typeenv.ml`, SATySFi does not check whether a type constructor has the same arity in the structure and the signature:
https://github.com/gfngfn/SATySFi/blob/69e62f66056b5644461f73c5fe841003602b0938/src/frontend/typeenv.ml#L1221

For example, SATySFi does not complain about the following code:

```
module M : sig
  type t
end = struct
  type 'a t = int
end
```

The type `t` expects 1 argument in the structure while in the signature it expects 0 arguments. 

Although this situation does not make anything unfavorable (e.g., breach of soundness), early error reports are preferable.

elpinal:
This PR refactors signature matching.

The current implementation of `sigcheck` highly depends on the syntax of signatures.
However, this situation makes the future development more difficult, in anticipation of the following mechanisms:

- Module renaming: `module M = N.O`
- Signature bindings: `module type S = ...`
- Module specifications: `module M : ...` in signatures
- Signature specifications: `module type S = ...` in signatures
- Transparent type specifications: `type t = ...` in signatures
- Functors: `module F (X : S) : T = ...`

Following [1], I introduced the notion of semantic signatures. I implemented signature matching in terms of semantic signatures, abandoning `sigcheck`.

## User-facing changes

These changes are not a complete no-op. This PR

- Fixes #193.
- Disallows duplicate specifications (requiring #192 to be merged; EDIT: It's been merged now).

Forbidding duplicate **type** specifications is reasonable because there is no valid way to find concrete definitions for multiple abstract type specifications for the same name. In contrast, whether allowing duplicate **value** specifications is just a matter of taste. Standard ML seems to disallow any kind of multiple specifications; OCaml seems to allow only multiple **value** specifications. I prefer Standard ML's way in favor of uniformity (and since SATySFi has no signature inheritance like OCaml's `include` yet, allowing duplicate specifications is for now not particularly useful).

## Technical notes

Following [2], I used *type locators*, implemented as the `location` field of `var_info` in `semanticSig.ml`.

Type locators are mappings from type variables (representing abstract types) to label sequences, making signature matching (especially `lookup_var` in `typeenv.ml`) more deterministic (than the implementation without type locators). Type locators are not necessarily needed to implement signature matching, but make the implementation more succinct and efficient.

Since I'm an OCaml beginner, the changes may be less sophisticated. Questions are welcome!

## References

- [1] Claudio V. Russo. Non-dependent types for Standard ML modules. 1999.
- [2] Derek Dreyer. A type system for recursive modules. 2007.


elpinal:
Carelessly, I've introduced a regression related to `open M` where `M` contains constructors. The third commit will fix it.

elpinal:
Since the new `SemanticSig` module does not depend on anything in `types.cppo.ml`, I separated it into another file named `semanticSig.ml`.

elpinal:
In my fork, I [implemented](https://github.com/elpinal/SATySFi/compare/refactor/semantic-signature...elpinal:signature-binding) signature bindings on top of this PR, thereby allowing programmers to write code like:

```ocaml
module type S = sig
  type t
  val x : t
end

module M : S = struct
  type t = int
  let x = 1
end
```

elpinal:
I found another regression, where the following code was rejected by the typechecker.

```ocaml
module M : sig
  type t

  val v : t option
end = struct
  type t = int

  let v = Some(4)
end
```

I fixed it by the latest commit.

elpinal:
In my fork, I [implemented](https://github.com/elpinal/SATySFi/compare/signature-binding...elpinal:transparent-type-specification) transparent type specifications on top of this PR. Thanks to this PR, implementing transparent type specifications (except variant types) was extremely easy.

Such specifications enable us to write code like the following.

```ocaml
module M : sig
  type t = int
  type u = C of t * u
end = struct
  type t = int
  type u = C of t * u
end
```

In this code, we can declare type equations in the signature, hence the signature language is made more expressive.

elpinal:
Again, I found a regression, where the following code was rejected: the handling of the types of `direct`-ed values was wrong. (Abstraction was not enforced for such values.)

```
module M : sig
  type t

  val make : int -> t

  direct \f : [t] inline-cmd
end = struct
  type t = int

  let make n = n

  let-inline \f n = (n + 1) |> arabic |> embed-string
end

let-inline \z = {\f(M.make 7);}
```

I fixed it by the latest (`0f612ed`) commit. (This commit was force-pushed because the commit message has grammatical errors.)

elpinal:
Conflicts have been resolved.

elpinal:
`ModuleInterpreter` has been renamed `ModuleClassifier` because the term "interpreter" gives one impression that it deals with the dynamic semantics of modules, while it really deals with mainly the static semantics. The term "classifier" is taken from "classification judgement" in [Russo 1999].

elpinal:
Now @gfngfn has been learned how to implement modules. Closing.

PR #195, opened 2019-09-23
Add OCaml 4.08 and 4.09 in Travis CI

XVilka:


XVilka:
I wonder why Travis is not triggered

nekketsuuu:
:eyes: Travis itself says the CI was triggered and passed: <https://travis-ci.org/gfngfn/SATySFi/builds/588330199>. So some connection between Travis and GitHub seems to be broken.

@gfngfn Did you change a configuration of GitHub in May 2019? According to a commit log, the connection was broken around that.

a log of git commits. In Mar 10, 2019, green chackmarks are shown besides commits. But they are not in May 18, 2019.

PR #197, opened 2019-10-05
Build some docs in CI to briefly check SATySFi works

KeenS:
This patch reinforces the CI to check some attached documents compile.
By doing this, you can briefly confirm following facts:

1. satysfi and some of its library correctly work
2. some of attached documents compiles


KeenS:
Ready to review :eyes:

PR #200, opened 2019-10-19
Add read external file support

puripuri2100:
This PR adds read external file support.

Added primitives

- `read-file : string -> string list`
Extract contents from file name.

Example

```
% This file's name is readfile.saty.
@require: stdjareport

document (|
  title = {test};
  author = {puripuri2100};
  show-title = true;
  show-toc = true;
|) '<
  +p(read-file `readfile.saty` |> List.fold-left (^) ` ` |> embed-string);
>
```
compile ==>
``% This file's name is readfile.saty.@require: stdjareportdocument (| title = {test}; author = {puripuri2100}; show-title =
true; show-toc = true;|) '< +p(read-file `readfile.saty` |> List.fold-left (^) ` ` |> embed-
string);>``

This primitive is useful for creating code examples.
I think it's safe because it only returns a string list.

### DIFF
--- a/src/frontend/bytecomp/vminstdef.yaml
+++ b/src/frontend/bytecomp/vminstdef.yaml
@@ -2622,3 +2622,33 @@ params:
 
 code: |
   raise (report_dynamic_error msg)
+
+---
+inst: ReadFile
+is-pdf-mode-primitive : yes
+is-text-mode-primitive : yes
+name: "read-file"
+type: |
+  ~% (tS @-> tL(tS))
+
+params:
+  - relpath : string
+
+code: |
+  let path =
+    Filename.concat (OptionState.job_directory ()) relpath
+  in
+  let str_list = ref [] in
+  let read filename =
+    let f = open_in filename in
+    try
+      while true do (str_list := (input_line f) :: !str_list) done
+    with End_of_file -> ()
+  in
+  let () =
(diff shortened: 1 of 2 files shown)


puripuri2100:
This commits add commands that takes a file name and returns code.

- `+file-code : [string] block-cmd`
- `\file-code : [string] inline-cmd`

And existing code changed slightly. There are no compatibility issue.

src/frontend/bytecomp/vminstdef.yaml
```diff
@@ -2622,3 +2622,33 @@ params:
...
+name: "read-file"
+type: |
+  ~% (tS @-> tL(tS))
+
+params:
+  - relpath : string
+
+code: |
+  let path =
+    Filename.concat (OptionState.job_directory ()) relpath
```
gfngfn: Is there a reason to support only relative paths and not to support absolute paths? IMHO it will be more convenient to allow absolute paths.

puripuri2100: It is difficult to determine which is better.
I think it's best to implement both.

gfngfn:
I’m sorry for the late reply… I merged this extension after adding to it an operation for asserting for security reasons that given relative paths do not contain `".."`.

PR #203, opened 2019-11-07
fix ipaexfont url

kamijin-fanta:
migrated to a new domain.


gfngfn:
I am sorry for the late response. I confirmed that the updated version works fine.

Thank you for amendment!

PR #206, opened 2019-12-08
Fix build failure caused by core_kernel.v0.13.0

amutake:
From `core_kernel.v0.13.0`, `Core_kernel.Heap` which is used in `src/backend/flowGraph.ml` is removed.
So I have restricted `core_kernel` version `< v0.13.0`.

This failure only happens with `ocaml {>= 4.08.0 & < 4.10.0}` because `core_kernel.v0.13.0` requires it.

(I noticed this by [nightly build of satysfi-docker](https://github.com/amutake/satysfi-docker/commit/676f7c5b91c2a8bab67e376ea33cf80fc7ac8c21/checks?check_suite_id=347370794))

### DIFF
--- a/satysfi.opam
+++ b/satysfi.opam
@@ -24,7 +24,7 @@ depends: [
   "batteries"
   "camlimages" {>= "5.0.1"}
   "camlpdf" {= "2.2.2+satysfi"}
-  "core_kernel" {>= "v0.10.0"}
+  "core_kernel" {>= "v0.10.0" & < "v0.13.0"}
   "cppo" {build & >= "1.6.4" & < "1.7.0"}
   "depext"
   "dune" {build}


na4zagin3:
It looks like that `Heap` has been renamed with `Paring_heap` since Core v0.13.
https://github.com/janestreet/core_kernel/blob/v0.13/pairing_heap/src/pairing_heap.mli

na4zagin3:
I just submitted #207 because I think development version (i.e., master branch) needs to follow upstream unless unless there is some particular reason.

amutake:
Thank you for the reply!

> I think development version (i.e., master branch) needs to follow upstream unless there is some particular reason.

I think so too, but I concern that SATySFi can only be built by only ocaml.4.08.* and 4.09.* because `core_kernel.v0.13.0` requires `ocaml >= 4.08.0 & < 4.10`...

na4zagin3:
That may be an unavoidable consequence. Is it acceptable to ask users who want to stick at OCaml 4.07.0 or earlier to use official releases (e.g., 0.0.3) or development version snapshots in maintained package repositories (e.g., satyrographos-repo)? Unlike JS or Python, it must be super easy for end-users to switch OCaml compiler versions.

Can we think this is a good chance of migrating to newer versions of OCaml? (Cf. CI with OCaml 4.08 and 4.09 https://github.com/gfngfn/SATySFi/pull/195)

I’d love to hear @gfngfn and others’ opinion.

na4zagin3:
Dune 2 series which has better handling of tests only supports OCaml 4.07 and newer. This could be another reason to drop support for at least OCaml 4.06.1.

amutake:
core_kernel.v0.13.1, which does not have version restriction of 'ocaml < 4.10.0', is released.
https://opam.ocaml.org/packages/core_kernel/core_kernel.v0.13.1/
I think we should use core_kernel.v0.13.1 or later, so I close this PR.

Thank you @na4zagin3 !

PR #207, opened 2019-12-08
Follow upstream change of Core

na4zagin3:
`Heap` has been renamed with `Pairing_heap` since Core v0.13.

Here is the summary of the discussion in https://github.com/gfngfn/SATySFi/pull/206.

There are two ways to workaround the incompatibility: 1) drop support of OCaml 4.07 and earlier or 2) stick at Core v0.12.

The second option implies SATySFi would be left behind development of OCaml forever. I am thus for the first option, i.e., supporting only 4.08 and later.

### DIFF
--- a/satysfi.opam
+++ b/satysfi.opam
@@ -24,7 +24,7 @@ depends: [
   "batteries"
   "camlimages" {>= "5.0.1"}
   "camlpdf" {= "2.2.2+satysfi"}
-  "core_kernel" {>= "v0.10.0"}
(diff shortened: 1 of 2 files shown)


na4zagin3:
It's found out that this fix does not work with core_kernel.v0.13.1. It's been fixed in the master https://github.com/gfngfn/SATySFi/commit/5ff61ee02ec4c397813e5b2b9c02a7573f5d83b3

Closing.

gfngfn:
Thank you all the same for the suggestion of improvement!

PR #212, opened 2020-01-23
Add --no-default-config option

na4zagin3:
This PR introduces a new option, `--no-default-config`, which keeps SATySFi from searching files from default configuration search paths.

This option `--no-default-config` will be used with `--config` option introduced by #161.

|Commandline Options|Search paths|
|---|---|
| (default) | `~/.satysfi`, `/usr/local/share/satysfi`, `/usr/share/satysfi` |
| `-C <paths>` | paths, `~/.satysfi`, `/usr/local/share/satysfi`, `/usr/share/satysfi` |
| `-C <paths> --no-default-config` | paths |

## Background

Library authors want to ensure if their library depend on only explicitly specified libraries. 

Currently, users or their build scripts need to manually update/clean-up files under the default search paths. This is not always possible. Non-sudo users cannot modify `/usr/...`. Build scripts in OPAM package cannot make any modification on `~/.satysfi` or `/usr/...`, either.

This can be useful for reproducible builds.

Cf. https://github.com/na4zagin3/satyrographos/issues/5

### DIFF
--- a/src/backend/optionState.ml
+++ b/src/backend/optionState.ml
@@ -19,6 +19,7 @@ type state = {
   mutable debug_show_block_space : bool;
   mutable mode             : (string list) option;
   mutable extra_config_paths : string list option;
+  mutable no_default_config_paths : bool;
 }
 
 
@@ -36,6 +37,7 @@ let state = {
   debug_show_block_space = false;
   mode             = None;
   extra_config_paths = None;
+  no_default_config_paths = false;
 }
 
 let set_input_kind ikd = state.input_kind <- ikd
@@ -85,3 +87,6 @@ let is_text_mode () =
 
 let set_extra_config_paths lst = state.extra_config_paths <- Some(lst)
 let get_extra_config_paths () = state.extra_config_paths
+
+let set_no_default_config_paths () = state.no_default_config_paths <- true
+let get_no_default_config_paths () = state.no_default_config_paths
--- a/src/backend/optionState.mli
+++ b/src/backend/optionState.mli
@@ -46,3 +46,6 @@ val is_text_mode : unit -> bool
 
 val set_extra_config_paths : string list -> unit
 val get_extra_config_paths : unit -> string list option
+
+val set_no_default_config_paths : unit -> unit
+val get_no_default_config_paths : unit -> bool
--- a/src/frontend/main.ml
+++ b/src/frontend/main.ml
@@ -433,7 +433,8 @@ let error_log_environment suspended =
   | NoLibraryRootDesignation ->
       report_error Interface [
         NormalLine("cannot determine where the SATySFi library root is;");
-        NormalLine("set appropriate environment variables.");
+        NormalLine("set appropriate environment variables");
+        NormalLine("or specify configuration search paths with -C option.");
       ]
 
   | NoInputFileDesignation ->
--- a/src/frontend/main.ml
+++ b/src/frontend/main.ml
@@ -956,6 +956,7 @@ let arg_spec_list curdir =
     ("--show-fonts"      , Arg.Unit(OptionState.set_show_fonts)      , " Displays all the available fonts"                      );
     ("-C"                , Arg.String(arg_config)                    , " Add colon-separated paths to configuration search path");
     ("--config"          , Arg.String(arg_config)                    , " Add colon-separated paths to configuration search path");
+    ("--no-default-config", Arg.Unit(OptionState.set_no_default_config_paths), "Do not use default configuration search path"         );
   ]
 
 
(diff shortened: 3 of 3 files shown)


nekketsuuu:
I don't understand why the `-C` option is not enough. Why do you want to ignore default paths? Modules can be used as namespaces. Is there a concrete example which illustrates the need for this option?

nekketsuuu:
> This option `--no-default-config` will be used with `--config` option introduced by #161.

Then it will be more user-friendly that SATySFi warns about using `--no-default-config` when used without `--config`/`-C`.

na4zagin3:
I have two examples.

Say, there are two libraries A and B where some packages in library A depends on B.

I wrote unit tests and build scripts for them. While they worked without errors in my local machine, in CI unit tests of library A failed due to lack of packages in library B.

I fact, the CI failed because the build script of A didn't contain dependency on B, which was expected. But it didn't failed in my local machine. Why?

I found out my local machine has library B installed under `/usr/local/share/satysfi/dist`.

How can test scripts prevent this problem? Here are some possible options which actually aren't.

- Can test scripts wipe out files under the default search paths? That is not a option considering they may run environment. For example, OPAM doesn't allow tests to modify those locations including `~/.satysfi`.
- Then, can test scripts parse all the packages and extract dependencies? That does not work well either. Parsing scripts would need to followup upstream grammar changes. And, it is impossible to statically analyze dynamic dependencies, for example @require statement produced with macros.

This is why I need a way to have SATySFi ignore the default search paths.

BTW, could you elaborate what do you mean by a module here?

> Modules can be used as namespaces.

na4zagin3:
> > This option --no-default-config will be used with --config option introduced by #161.
>
> Then it will be more user-friendly that SATySFi warns about using `--no-default-config` when used without `--config`/`-C`.

Good catch. How about updating an error message of `NoLibraryRootDesignation` error?

https://github.com/gfngfn/SATySFi/blob/65851d34a599cd7bbc03583162fc2a68160151ff/src/frontend/main.ml#L433-L437

nekketsuuu:
@na4zagin3 I supposed your problem may be related to a name collision. So I suggest to use modules. But now I consider your problem is originated from your design for testing.

First please note that your CI failed correctly and your local tests succeeded correctly, too. This is not a problem of (current) SATySFi, which does not manage dependencies by itself. This isn't a bug of test scripts, too. If this is a problem, it can be a issue of a package manager (if it's used). And also, an issue caused by manually-installed libraries is not supported by the package manager; that is your problem.

By the way, if you want a clean local environment for SATySFi, then I personally prefer virtualenv-like software over the `--no-default-config` option because you may need a custom environment next. How about implementing a feature which is similar to `opam env` or [Python's several tools](https://stackoverflow.com/a/41573588/5989200)? For example, how about generalize [`SATYSFI_RUNTIME`](https://github.com/gfngfn/SATySFi/blob/69e62f66056b5644461f73c5fe841003602b0938/src/frontend/main.ml#L962) introduced by #67?

na4zagin3:
Thank you for your summarizing.

Well, as you said, the very problem with the current SATySFi implementation is that it's impossible for non privileged users to create a clean SATySFi environment. Currently it's not possible to write a package manager running as a _non-priviledged_ user which have SATySFi ignore system- or user-installes packages.

Anyway, generalized SATYSFI_RUNTIME was what I firstly planned and am still preferring over current solution with `-C` option. Not only Python, but most of combat-proven tools do use environment variables like that.

nekketsuuu:
@gfngfn Any thoughts on this?

na4zagin3:
Let me add a few more. Of course I was examining generalized SATYSFI_RUNTIME, but yet I haven’t come up with a way to extend `SATYSFI_RUNTIME` backward compatibly. Here are sources of search paths.

- `extra_dirs`: paths specified with `-C` option
- `home_dirs`: only `~/.satysfi`
- `runtime_dirs`: `$SATYSFI_RUNTIME` on Win32, `["/usr/local/share/satysfi"; "/usr/share/satysfi"]` otherwise

What we need to create environment switcher like pyenv or opam is the way to empty `home_dirs` and `runtime_dirs`.

If we naively extend so that `SATYSFI_RUNTIME` overrides both `home_dirs` and `runtime_dirs`, it breaks in an environment where satysfi-dist is located under `~/.satysfi`. We could update the error message to tell users to add `~/.satysfi` to `SATYSFI_RUNTIME` when no paths are set.

Do you have any thoughts on that?

gfngfn:
I’m so sorry for the very late response…

Since it seems to me that `--no-default-config` option will be a convenient and concise means for detecting a cause of environment-related problems, I’m ready to merge this PR.

PR #215, opened 2020-02-06
Update download-fonts.sh

Niccori250k:
取得するIPAexフォントのバージョンをVer.003.01からVer.004.01に変更


nekketsuuu:
Nice, Reiwa has come :)

https://ipafont.ipa.go.jp/node21

> Changes from IPAex Fonts Ver.003.01.
> U+32FF “SQUARE ERA NAME REIWA” was added.

gfngfn:
Confirmed that the modified URI is correct. Thank you for improvement!

PR #218, opened 2020-02-17
Add a new primitive: linear-transform-path

monaqa:
This PR adds a new primitive
`linear-transform-path: float -> float -> float -> float -> path -> path`.
`path-origin |> linear-transform-path a b c d` is evaluated to a path into which the`path-origin` is transformed with the following formula:

Combining `linear-transform-path` with `shift-path`, you can implement any path-to-path affine transform.

This PR also provides two functions, `rotate-path` and `scale-path`,  in `gr` package.

### DIFF
--- a/lib-satysfi/dist/packages/gr.satyh
+++ b/lib-satysfi/dist/packages/gr.satyh
@@ -19,6 +19,8 @@ module Gr : sig
   val text-rightward : point -> inline-boxes -> graphics
   val arrow : length -> color -> length -> length -> length -> point -> point -> graphics list
   val dashed-arrow : length -> length * length * length -> color -> length -> length -> length -> point -> point -> graphics list
+  val rotate-path : point -> float -> path -> path
+  val scale-path : point -> float -> float -> path -> path
 
 end = struct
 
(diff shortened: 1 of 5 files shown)


gfngfn:
I’m so sorry for the late response. Confirmed that the added primitive and the functions work (as shown below). Thank you for enhancement!

```
@require: stdjareport
@require: gr

let-inline ctx \test deg =
  let width = 1cm in
  let height = 3cm in
  inline-graphics width height 0pt (fun ((x, y) as pt) -> (
    let path = Gr.polygon pt [(x +' width, y); (x +' width, y +' height)] in
    let center = (x +' width *' 0.5, y +' height *' 0.5) in
    let path-rotated = Gr.rotate-path center deg path in
    [
      stroke 0.5pt Color.black path;
      stroke 1pt Color.red path-rotated;
    ]
  ))
in

document (|
  title = {Linear Transformation by \@monaqa};
  author = {gfn};
  show-title = true;
  show-toc = false;
|) '<
  +p{
    ${30^\circ}: \test(30.0);
  }
  +p{
    ${60^\circ}: \test(60.0);
  }
  +p{
    ${120^\circ}: \test(120.0);
  }
  +p{
    ${300^\circ}: \test(300.0);
  }
>
```

rotation

PR #221, opened 2020-03-27
Change OCaml version into 4.10.0 of .travis.yml

y-yu:
- I was trying to add OCaml 4.10 testing on Travis CI
    - ~~but it failed...... 😇~~
    - It seems to be success! https://travis-ci.org/github/y-yu/Macrodown/builds/674997020
- It requires to update `camlpdf` to verison 2.3.1
    - so this PR depends on https://github.com/gfngfn/camlpdf/pull/3
    - 👆 was merged. Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,7 @@ os:
   - linux
   - osx
 env:
-  - OCAML_VERSION=4.06
-  - OCAML_VERSION=4.07
+  - OCAML_VERSION=4.10
 sudo: required
 install:
   - wget https://raw.githubusercontent.com/ocaml/ocaml-travisci-skeleton/master/.travis-ocaml.sh


na4zagin3:
https://github.com/gfngfn/SATySFi/pull/207 may be required to build with OCaml 4.10

y-yu:
@na4zagin3 

I tried to compile SATySFi with https://github.com/gfngfn/camlpdf/pull/3 in Travis CI but it failed 😇 

- https://travis-ci.org/github/y-yu/Macrodown/jobs/668007137#L672

 I don't know whether the error 👆 is what you said or not....

na4zagin3:
@y-yu It's different problem from that error.

The error you posted is related to the compile option specified at `src/dune`:
https://github.com/gfngfn/SATySFi/blob/84007341a12808bca74819a055ab5e55cbc0b285/src/dune#L4

Option `-unsafe-string` specified here has been removed (https://github.com/ocaml/ocaml/pull/1859).

y-yu:
@na4zagin3 sorry for my late reply 🙇 

I merged the change on the `master` branch and run CI job.

- https://travis-ci.org/github/y-yu/Macrodown/jobs/674967132
- image

Indeed SATySFi can be built by OCaml4.10 🎉 but it discontinued to support other OCaml versions....

PS:
Who check the CI result of `master` branch? It seems that `master` branch haven't been built by all OCaml versions except 4.10 already 🤔

y-yu:
For now I added to the setting to `.travis.yml` to ignore failures in OCaml 4.06 and 4.07.

image

na4zagin3:
@y-yu  Can you _remove_ builds against OCaml 4.06 and 4.07 as they are no longer supported?

https://github.com/gfngfn/SATySFi/blob/master/satysfi.opam#L23

y-yu:
OK. > remove builds against OCaml 4.06 and 4.07

I agree with that because these versions will never be supported again.

gfngfn:
Seen from Travis CI (which has been set up thanks to #224), this PR seems working well. Thank you for the improvement!

Issue #222, opened 2020-03-27
Build error with core_kernel v0.13.1


yasuo-ozu:
Error msg:

```shell
# dune build
#       menhir src/parser.{ml,mli}
# Warning: 2 states have shift/reduce conflicts.
# Warning: 2 shift/reduce conflicts were arbitrarily resolved.
# File "src/backend/flowGraph.ml", line 8, characters 14-30:
# Error: Unbound module Core_kernel.Heap
```

source:

```src/backend/flowGraph.ml
module Heap = Core_kernel.Heap
```

na4zagin3:
Duplicated: #206 & #207

STNeeic:
I've renamed `Core_kernel.Heap` to `Core_kernel.Pairing_heap` but still failed to build.

Error msg:

```
# dune build
#       menhir src/parser.{ml,mli}
...
# File "src/backend/flowGraph.ml", line 8, characters 14-38:
# Error: Unbound module Core_kernel.Pairing_heap
```

versions info
```
# Packages matching: installed
# Package                     # Installed   # Synopsis
base.v0.13.1                  v0.13.1       Full standard library replacement for OCaml
...
core_kernel.v0.13.1           v0.13.1       Industrial strength alternative to OCaml's standard library
...
ocaml.4.09.0                  4.09.0        The OCaml compiler (virtual package)
ocaml-base-compiler.4.09.0    4.09.0        Official release 4.09.0
```

na4zagin3:
@STNeeic You're correct. core_kernel v0.13.1 includes further changes (see https://github.com/gfngfn/SATySFi/commit/5ff61ee02ec4c397813e5b2b9c02a7573f5d83b3).

STNeeic:
Successfully installed. Thx All!

PR #223, opened 2020-03-27
limit core_kernel version and use --safe-string

yasuo-ozu:
This update depends on camlpdf.2.3.1. Please update gfngfn/satysfi-external-repo and gfngfn/camlpdf.


yasuo-ozu:
Issue #222

na4zagin3:
Partially duplicated: #206 & #207

gfngfn:
Since it was a good opportunity, I chose to use `core_kernel >= v0.13` (which has been suggested by #207, as @na4zagin3 noted above).

PR #226, opened 2020-03-28
Move file generation rules in Makefile into Dune files

### LINKED ISSUE #198: Duplicated VM instruction definition files
Currently, there are 2 VM instruction definition files in this repository:
* src/frontend/bytecomp/vminstdef.yaml
* tools/gencode/vminst.ml

This may confuse contributors who wants to add new primitives. According to the Makefile, the latter is generated by the former. So, I think it should be removed from this repository. I also think the name "vminstdef" should be like "primitivedef" because it is also used by the interpreter and it only contains primitive definitions now.

na4zagin3:
This PR moves file generation rules into Dune from Makefile.

`make all`, `make install`, and `make uninstall` still work as before, although I think they are no longer useful. Let me know if I need to remove the Makefile.

Fixes #198


na4zagin3:
Well, I just realized this PR is similar to https://github.com/gfngfn/SATySFi/pull/142, except that this PR does not remove dependency on Ruby.

na4zagin3:
Do you want to replace vminst.ml with vminstdef.yaml or the other way around?

TonalidadeHidrica:
Is this PR a superset of #142?

gfngfn:
I’m so sorry for the extremely late response… Thank you for trying to improve the build process!

> Do you want to replace vminst.ml with vminstdef.yaml or the other way around?

The latter seems preferable to me. The file `vminst.ml` was originally introduced by #99 as an intermediate step for removing the build dependency on Ruby.

na4zagin3:
@gfngfn Thank you for your reply!  Then, I'm closing this PR in favor of #142.

gfngfn:
Thank you all the same for trying to improve the build process!

PR #228, opened 2020-04-21
Fix IPAexfont version in download-fonts.sh

amutake:
There was an omission of #215.

(I think #197 will make us realize these mistakes!)


gfngfn:
Ah, I missed that! Thank you.

amutake:
Thank you!

PR #229, opened 2020-04-24
Fix OCaml version on README

puripuri2100:


puripuri2100:
Now, OCaml version on README is 4.06.0, but actually >= 4.08.0 is necessary.
So I OCaml rewritten version on README to 4.10.0

y-yu:
LGTM

gfngfn:
Thanks for improvement!

PR #230, opened 2020-04-29
Update SHA1 hash file for the IPA Font

na4zagin3:
This PR follows up https://github.com/gfngfn/SATySFi/pull/215 and https://github.com/gfngfn/SATySFi/pull/228 which did not update the hash file.

This problem was found during investigation of CI failures of <https://github.com/na4zagin3/satyrographos-repo/pull/110>.

PR #231, opened 2020-04-29
Refactor download-fonts.sh

na4zagin3:
This PR depends on #230. It refactors download-fonts.sh with

- setting `set -ue` to reduce `|| exit 1` and have more unconsidered cases (i.e., #230) fail,
- and introducing functions `show_message`, `validate_file`, and `download_file`.

I used subshell `( ... )` to work around `local` missing in POSIX. If you are fine with using `/bin/bash`, I would like to rewrite with more conventional idioms.


gfngfn:
Confirmed that the modified version of `download-fonts.sh` works. Thank you for improvement!

PR #234, opened 2020-05-23
Fix: suppress page break just after +section commands

monaqa:
This PR fixes that the page break does not appear right after the chapter/section title.

# Example

The second and third page of `doc/doc-primitives.saty` is shown as an example.

---
Before:
before

---
After:
after

---


gfngfn:
Thank you for fixing!

PR #240, opened 2020-10-01
Fix the URL of IPAexfont

amutake:
The URL of IPAexfont has been changed.
see: https://ipafont.ipa.go.jp/


gfngfn:
Thank you for the update!

PR #241, opened 2020-10-03
Add a new primitive: linear-transform-graphics

monaqa:
(graphics version of #218.)

This PR adds a new primitive
`linear-transform-graphics: float -> float -> float -> float -> graphics -> graphics`.
`gr-origin |> linear-transform-graphics a b c d` is evaluated to a graphics value into which the `gr-origin` is transformed with the following formula:

Combining `linear-transform-graphics` with `shift-graphics`, you can implement any graphics-to-graphics affine transform (rotation, scaling, reflection, etc.).

This PR also provides two functions, `rotate-graphics` and `scale-graphics`,  in `gr` package.

There are two differences between `linear-transform-path` and `linear-transform-graphics`:

* `linear-transform-graphics` can transform any inline-boxes using `draw-text` primitive.
* When you transform a stroke with `linear-transform-graphics` command, the stroke thickness is changed (unless a congruence transformation).

### DIFF
--- a/lib-satysfi/dist/packages/gr.satyh
+++ b/lib-satysfi/dist/packages/gr.satyh
@@ -21,6 +21,8 @@ module Gr : sig
   val dashed-arrow : length -> length * length * length -> color -> length -> length -> length -> point -> point -> graphics list
   val rotate-path : point -> float -> path -> path
   val scale-path : point -> float -> float -> path -> path
+  val rotate-graphics : point -> float -> graphics -> graphics
+  val scale-graphics : point -> float -> float -> graphics -> graphics
 
 end = struct
 
@@ -184,5 +186,20 @@ end = struct
     path |> shift-path (0pt -' centx, 0pt -' centy)
          |> linear-transform-path scalex 0. 0. scaley
          |> shift-path centpt
+
+
+  let rotate-graphics centpt angle gr =
+    let (centx, centy) = centpt in
+    let rad = angle *. math-pi /. 180. in
+    gr |> shift-graphics (0pt -' centx, 0pt -' centy)
+       |> linear-transform-graphics (cos rad) (0. -. (sin rad)) (sin rad) (cos rad)
+       |> shift-graphics centpt
+
+
+  let scale-graphics centpt scalex scaley gr =
+    let (centx, centy) = centpt in
+    gr |> shift-graphics (0pt -' centx, 0pt -' centy)
+       |> linear-transform-graphics scalex 0. 0. scaley
+       |> shift-graphics centpt
 
 end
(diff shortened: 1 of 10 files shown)


monaqa:
Code example:
```
@require: stdjareport
@require: gr

let-inline ctx \test deg =
  let width = 1cm in
  let height = 3cm in
  inline-graphics width height 0pt (fun ((x, y) as pt) -> (
    let path = Gr.polygon pt [(x +' width, y); (x +' width, y +' height)] in
    let center = (x +' width *' 0.5, y +' height *' 0.5) in
    [
      stroke 0.5pt Color.black path;
      stroke 1pt Color.red path |> Gr.rotate-graphics center deg;
    ]
  ))
in

document (|
  title = {Linear Transformation of graphics};
  author = {\@monaqa};
  show-title = true;
  show-toc = false;
|) '<
  +p{
    ${30^\circ}: \test(30.0);
  }
  +p{
    ${60^\circ}: \test(60.0);
  }
  +p{
    ${120^\circ}: \test(120.0);
  }
  +p{
    ${300^\circ}: \test(300.0);
  }
>
```

Result:
linear-transform-graphics-demo

(Since rotation is part of the congruent transformation, the result appears to be the same as when using `linear-transform-path`.)

monaqa:
I forgot to mention this change: https://github.com/gfngfn/SATySFi/pull/241/files#diff-ff6ba46343bcea35973ad4d439f43e94L217-R261

In the previous code, the coordinates specified by the cm command are applied to all subsequent text (because the cm command precedes the q command). So I switched the order of the two commands.
Since the cm command currently performs only identity transformation, discarding this changes will not cause any problems.

gfngfn:
Looks very well-written, and I confirmed that the modified implementation indeed works as expected! (especially the composition of transformations and the rotation of texts)

```
@require: stdjareport
@require: gr

let-inline ctx \test deg =
  let width = 1cm in
  let height = 3cm in
  inline-graphics width height 0pt (fun ((x, y) as pt) -> (
    let path = Gr.polygon pt [(x +' width, y); (x +' width, y +' height)] in
    let center = (x +' width *' 0.5, y +' height *' 0.5) in
    let gr-text = draw-text (x, y) (read-inline ctx {あいう}) in
    [
      stroke 0.5pt Color.black path;
      stroke 1pt Color.red path |> Gr.rotate-graphics center deg;
      stroke 1pt Color.blue path |> Gr.rotate-graphics center deg |> Gr.rotate-graphics center deg;
      gr-text;
      gr-text |> Gr.rotate-graphics center deg;
      gr-text |> Gr.rotate-graphics center deg |> Gr.rotate-graphics center deg;
    ]
  ))
in

document (|
  title = {Linear Transformation of graphics};
  author = {\@monaqa};
  show-title = true;
  show-toc = false;
|) '<
  +p{${30^\circ}: \test(30.0);}
  +p{${60^\circ}: \test(60.0);}
  +p{${120^\circ}: \test(120.0);}
  +p{${300^\circ}: \test(300.0);}
>
```

スクリーンショット 2020-12-14 1 18 25

Thank you so much for enhancement!

Issue #224, opened 2020-03-28
Is Travis CI testing NOT working?


y-yu:
- Travis CI status has not been shown 🤔 
- ![image](https://user-images.githubusercontent.com/612043/77818026-94526800-7112-11ea-88b6-e98947203ea1.png)

y-yu:
I encountered the same thing in https://github.com/rust-lang-ja/book-ja-pdf/pull/5 . And it was solved by this way 👇 

1. Access to the [GitHub applications admin page](https://github.com/settings/applications)
    - And look at `Authorized OAuth Apps` tab
2. Look for the `Travis CI for Open Source` in the tab
    - ![image](https://user-images.githubusercontent.com/612043/79068763-3a51c500-7d04-11ea-8316-6ebf5a960869.png)
3. Revoke the `Travis CI for Open Source`
    - <img src="https://user-images.githubusercontent.com/612043/79068771-52294900-7d04-11ea-882c-3bb7ecaa4362.png" width="40%"/>
4. Access to https://travis-ci.org/ (NOT travis-ci.**com**) and sign out if you are signed in
    - <img src="https://user-images.githubusercontent.com/612043/79068811-a03e4c80-7d04-11ea-8869-a8f85eae2d4a.png" width="40%"/>
5. Sign in the circle-ci.org by your GitHub account
6. Authorize Circle CI for GitHub again

Good luck!

gfngfn:
I’m sorry for the late response. I performed just now the procedure you had kindly introduced to me!

gfngfn:
Is this sufficient to run CI next time when pushing commits to this repository? I mean I wonder whether I still have another task like activating the connection of this repository and Circle CI.

y-yu:
Sorry, I wrote `circle-ci.org`. It was a mistake!!! I insisted `travis-ci.org`.

gfngfn:
I understand. I signed in `travis-ci.org` just now.

gfngfn:
The CI seems working well. Thank you very much for your kind introduction!

y-yu:
CI is running like that 👉 https://travis-ci.org/github/gfngfn/SATySFi/builds/677166453 but the results are *not* shown from GitHub.... 😇 

![image](https://user-images.githubusercontent.com/612043/79741544-67255e00-833c-11ea-9382-89b46c3e371c.png)

[...]

PR #250, opened 2020-11-23
Add GitHub Actions

y-yu:
- Travis CI( [travis-ci.**com**](https://travis-ci.com) ) has been required an application to use their resource unlimitedly even though we are OSS works
    - https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
    - > We will be offering an allotment of OSS minutes that will be reviewed and allocated on a case by case basis. Should you want to apply for these credits please open a request with Travis CI support stating that you’d like to be considered for the OSS allotment
    - See also: [今までありがとうTravis CI、さよならTravis CI - xuwei-k's blog](https://xuwei-k.hatenablog.com/entry/2020/11/04/161400)
- We use [travis-ci.**org**](https://travis-ci.org) so it's not irrelevant for now.....
    - But [travis-ci.org](https://travis-ci.org) is now _deprecated_ 😢 
    - Since [travis-ci.org](https://travis-ci.org) is using GitHub V1 API, it causes many problems such like the build status would not be updated
    - In the common sense, we have to move [travis-ci.com](https://travis-ci.com) even if we keep using Travis CI
        - https://github.com/gfngfn/SATySFi/issues/224#issuecomment-616461196
- Then I make GitHub Action setting, which could be the next our CI service
- Travis CI build test takes 40-50minutes, GitHub Actions seems to be faster than Travis CI
    - 
    - On the other hand, GitHub Actions CI took 16 minutes
        - Example result: https://github.com/y-yu/Macrodown/actions/runs/375810549
        - 
- So it should be better to use GitHub Actions instead of Travis CI, shouldn't it? 🤔

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,34 @@
+name: CI
+
+on:
+  [pull_request, push]
+
+jobs:
+  build:
+    name: Build
+    strategy:
+      matrix:
+        os: [ 'ubuntu-latest', 'macos-latest' ]
+        ocaml-version:
+          - 4.10.1
+          - 4.11.1
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Setup OCaml ${{ matrix.ocaml-version }}
+        uses: avsm/setup-ocaml@v1
+        with:
+          ocaml-version: ${{ matrix.ocaml-version }}
+        
+      - name: Install SATySFi dependencies
+        run: |
+          opam repository add satysfi-external https://github.com/gfngfn/satysfi-external-repo.git
+          opam update
+          opam pin add satysfi.dev . --no-action
+          opam depext satysfi --yes --with-doc --with-test
+          opam install . --deps-only --with-doc --with-test
+
+      - name: Build SATySFi
+        run: opam exec -- make all


.github/workflows/ci.yml
```diff
@@ -0,0 +1,35 @@
...
+        uses: avsm/setup-ocaml@v1
+        with:
+          ocaml-version: ${{ matrix.ocaml-version }}
+        
+      - name: Install SATySFi dependencies
+        run: |
+          opam repository add satysfi-external https://github.com/gfngfn/satysfi-external-repo.git
+          opam update
+          opam pin add -y --no-action satysfi .
+          opam install -y --deps-only satysfi
```
smorimoto, line 30:
```suggestion
          opam repository add satysfi-external https://github.com/gfngfn/satysfi-external-repo.git
          opam update
          opam pin add satysfi.dev . --no-action
          opam depext satysfi --yes --with-doc --with-test
          opam install . --deps-only --with-doc --with-test
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: Build SATySFi
+        run: |
+          eval `opam config env`
+          make all
```
smorimoto, line 35:
```suggestion
        run: opam exec -- make all
```

smorimoto:
Looks good apart from minor nits.

.github/workflows/ci.yml
y-yu: I don't know OCaml CI fashion very well 😇  
Do @na4zagin3 and @gfngfn think it's better than the current one?

smorimoto: It depends what you mean by better.

smorimoto: Except for the depext part, there are no other major and/or minor changes. (See here for a description of what depext is.)
https://github.com/avsm/setup-ocaml#what-is-the-difference-between-opam-dependencies-and-depext-dependencies
Note that GitHub Actions VM images are relatively unstable. With that in mind, even if depext is not currently needed, it makes sense to include it as part of our workflows.

y-yu: OK, thanks for your explanation to me 🙇 
I'll follow your suggestion and rebase these commit.

smorimoto: I work with OCaml Labs and GitHub to work on Actions itself and a lot of infrastructure stuff like setup-ocaml, so feel free to ping me if you have any questions :)

y-yu:
@gfngfn
Please review or comment to this PR 🙏

smorimoto:
Btw is this really a huge rush?

gfngfn:
I’m so sorry for the late response:pray:… Let me take a close look into this PR tonight.

y-yu:
Thank you for your reply. As smorimoto san said, it would be OK to be late for merging/close this PR. I concerned that this PR will fade away 🙃

gfngfn:
I understand what’s described on `.github/workflows/ci.yml`.

I pushed the last commit and invoke the pipeline just to be sure, and then confirmed that it seems working fine! https://github.com/gfngfn/SATySFi/runs/1590159714

Thank you so much for setting up GitHub Actions!

Issue #252, opened 2020-12-05
sha1sum does not exist on Mac


na4zagin3:
[`./download-fonts.sh`](https://github.com/gfngfn/SATySFi/blob/c841df2d68738f1d5cd86d59e14e7bbd4137d8d8/download-fonts.sh#L7) depends on `sha1sum`, while it's not available on Mac.

This problem causes https://github.com/monaqa/slydifi/issues/15.

Here is the execution result of `./download-fonts.sh` on Mac.  As you can see, `./download-fonts.sh` does download files at the second run because file verification fails (Thank you, @monaqa)

```
[2020-12-05 16:11:30]
SATySFi on  master via 🐫 v4.10.0 via 💎 v2.6.3
❯ sh -x ./download-fonts.sh
+ set -ue
+ CACHE=temp
+ MESSAGE_PREFIX='[download-fonts.sh]'
+ SHA1SUM=sha1sum
++ dirname ./download-fonts.sh
+ cd .
+ mkdir -p temp
+ NAME=lm2.004otf
+ download_file lm2.004otf.zip http://www.gust.org.pl/projects/e-foundry/latin-modern/download/lm2.004otf.zip
+ NAME=lm2.004otf.zip
+ URL=http://www.gust.org.pl/projects/e-foundry/latin-modern/download/lm2.004otf.zip
+ '[' -f temp/lm2.004otf.zip ']'
+ validate_file lm2.004otf.zip
+ NAME=lm2.004otf.zip
+ cd temp
+ sha1sum --check lm2.004otf.zip.sha1
./download-fonts.sh: line 20: sha1sum: command not found
+ show_message 'downloading '\''lm2.004otf.zip'\'' ...'
+ echo '[download-fonts.sh] downloading '\''lm2.004otf.zip'\'' ....'
[download-fonts.sh] downloading 'lm2.004otf.zip' ....
+ wget -O temp/lm2.004otf.zip http://www.gust.org.pl/projects/e-foundry/latin-modern/download/lm2.004otf.zip
--2020-12-05 16:11:35--  http://www.gust.org.pl/projects/e-foundry/latin-modern/download/lm2.004otf.zip
www.gust.org.pl (www.gust.org.pl) をDNSに問いあわせています... 158.75.62.5
www.gust.org.pl (www.gust.org.pl)|158.75.62.5|:80 に接続しています... 接続しました。
HTTP による接続要求を送信しました、応答を待っています... 200 OK
[...]

PR #253, opened 2020-12-05
Detect SHA checksum utilities

na4zagin3:
This commit has `download-fonts.sh` use `shasum` or `sha1sum` in this order and fail if neither of them exists.

Closes https://github.com/gfngfn/SATySFi/issues/252


gfngfn:
Thanks for inprovement!

PR #258, opened 2021-01-23
Allow setting LIBDIR from commandline argument

na4zagin3:
This commit has install-libs.sh take an optional argument `LIBDIR`.


gfngfn:
Thanks!

PR #263, opened 2021-02-16
update ocaml version in satysfi.opam

massn:
OCaml `4.09.1` doesn't work. 

```
$ ocaml --version
The OCaml toplevel, version 4.09.1
$ opam pin add satysfi .
...
The following dependencies couldn't be met:
  - satysfi → camlpdf >= 2.3.1+satysfi
      no matching version

[NOTE] Pinning command successful, but your installed packages may be out of sync.
```


gfngfn:
Thank you for fixing!

Issue #267, opened 2021-04-18
I want to be able to set up a document information dictionary.


puripuri2100:
Now, there is no information dictionary in the PDF file output by SATySFi.

<img width="560" alt="fig" src="https://user-images.githubusercontent.com/36466996/115134482-6a385480-a04b-11eb-9c4b-a91f8778f6e2.PNG">

I want to be able to set up a document information dictionary.

I'm thinking of code like this:

```ocaml
let () = set-information-dictionary (|
  title = Some(`title`);
  subjects = Some(`subject`);
  keywords = [`keyword`];
  author = Some(`author`);
|)
```

gfngfn:
I understand what you want to do. It seems worth implementing to me as well :+1:

According to *PDF Reference sixth edition*, Sections 3.4.4 and 10.2.1, the document information dictionary is supposed to be contained in the trailer dictionary by using `/Info` key. Based on the API of `camlpdf`, we can realize this functionality by modifying `trailerdict` field of values of type `Pdf.t`.

The formalization of `set-information-dictionary` looks, however, too imperative to me. Intuitively, an optional argument of page-breaking primitives seems to be a suitable place for such settings:

```
val page-break : ?(| title : string option; … |) -> …
```

PR #268, opened 2021-04-19
set up a document information dictionary

puripuri2100:
Solve #267 .

### DIFF
--- a/src/backend/documentInfomationDictionary.ml
+++ b/src/backend/documentInfomationDictionary.ml
@@ -45,7 +45,9 @@ let add_to_pdf pdf =
       | [s] -> str^s
       | s::xs -> sub xs (str^s^" ")
     in
-    [("/Keywords", Pdf.String (to_utf16be (sub lst "")))]
+    match lst with
+    | [] -> []
+    | lst -> [("/Keywords", Pdf.String (to_utf16be (sub lst "")))]
   in
   let creator =
     [
--- a/src/backend/documentInfomationDictionary.ml
+++ b/src/backend/documentInfomationDictionary.ml
@@ -18,22 +18,24 @@ let register dictionary =
   registered_document_infomation_dictionary := dictionary
 
 
+let to_utf16be text = InternalText.to_utf16be (InternalText.of_utf8 text)
+
 let add_to_pdf pdf =
   let document_infomation_dictionary = !registered_document_infomation_dictionary in
   let title =
     match document_infomation_dictionary.title with
     | None -> []
-    | Some(title) -> [("/Title", Pdf.String title)]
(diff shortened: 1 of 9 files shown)


gfngfn:
Thank you for suggesting modification! Let me look into the implementation.

src/backend/documentInfomationDictionary.ml
```diff
@@ -0,0 +1,67 @@
+type document_infomation_dictionary = {
```
gfngfn, line 1: Found a mis-spelling as an English word (`infomation` → `information`).

gfngfn (changes requested):

src/backend/documentInfomationDictionary.ml
```diff
@@ -0,0 +1,67 @@
...
+    let lst = document_infomation_dictionary.keywords in
+    let rec sub lst str =
+      match lst with
+      | [] -> str
+      | [s] -> str^s
+      | s::xs -> sub xs (str^s^" ")
+    in
+    match lst with
+    | [] -> []
+    | lst -> [("/Keywords", Pdf.String (to_utf16be (sub lst "")))]
```
gfngfn, line 50: I think this can be done by `[("/Keywords", Pdf.String(to_utf16be (String.concat " " lst)))]` without using `sub`.

gfngfn (changes requested):

puripuri2100:
I fixed.

gfngfn:
On a second thought during the trial to extend `page-break` with an optional argument, I have come to think that providing `set-document-information`, which you had first suggested, would be safer. The current formalization of how to handle primitives is very complicated, and thus it seems difficult and error-prone to extend primitives with optional arguments (although this situation should be solved as a distinct issue).

I am sorry for asking so much of you, but would you add `set-document-information` to `vminstdef.yaml` instead of extending `page-break`? (If not, I’m ready to do this by myself, of course.)

puripuri2100:
I added `set-document-information` to `vminstdef.yaml`.

puripuri2100:
Hmm...
Should I rename it to `register-document-information`, like `register-outline`?

gfngfn:
Thank you for the modification!

> Should I rename it to `register-document-information`, like `register-outline`?

It sounds nice! Both are acceptable to me, but I slightly prefer `register-*` to `set-*`. Feel free to rename the primitive if you will.

puripuri2100:
It's all done!
Please merge at any time.

gfngfn:
Confirmed that the primitive works fine (by using it in `stdjabook.satyh` and `doc-primitives.saty`)!

document information dictionary was successfully written in a PDF generated by SATySFi

Thank you for the development!

PR #271, opened 2021-04-26
Recommend Satyrographos for non-dev users

na4zagin3:
This PR replaces section “Install using Homebrew” with “Install using Satyrographos” to solve the following problems:

1. Due to the directory hierarchy of M1 Mac, `-C` option is required for the casual users that install SATySFi with Homebrew, which I think is not ideal. (See https://github.com/gfngfn/SATySFi/pull/269 for more details)
2. Users need to clone the git repo to install historical versions of SATySFi
3. Users need to manually copy files to install third-party libraries

Does this PR look okay, @nyuichi?  I'd love to restore the secton “Install using Homebrew” with a special mention for M1 Mac and Linuxbrew if you would like.

This partially addresses https://github.com/gfngfn/SATySFi/issues/265, in other words, this PR does not add a method with @amutake’s docker-satysfi because I'm not strongly sure which image we should recommend.

### DIFF
--- a/README-ja.md
+++ b/README-ja.md
@@ -17,12 +17,35 @@
 
 また，2021年2月現在も発展を続けています。
 
-## Homebrew を使ったインストール方法 (Mac ユーザ向け)
+## Satyrographos を使ったインストール方法 (初心者向け)
 
-Homebrew のフォーミュラが用意されています。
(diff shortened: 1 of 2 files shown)


gfngfn:
Sorry for the late notice. Thank you for adding explanations!

PR #276, opened 2021-06-18
Enhance Ocaml-related gitignore based on gitignore.io

TonalidadeHidrica:
Added some entries to `.gitignore` that are related to Ocaml, based on the recommendation by [gitignore.io](https://www.toptal.com/developers/gitignore).  One motivation is that I wanted to use local `_opam` switch, but it was not ignored by default.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,12 +2,7 @@
 \#*
 .\#*
 *~
-*.cmi
-*.cmo
-*.cmx
-*.o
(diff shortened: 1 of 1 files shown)


TonalidadeHidrica:
I'm sure it can be reviewed in a couple of minutes!

gfngfn:
I understand the motivation. Thank you!

PR #280, opened 2021-07-03
Add `clip-graphics-by-path` primitive

yasuo-ozu:
```ml
clip-graphics-by-path : path -> graphics -> graphics
```

This new primitive allows users to clip any graphics with capricious path.


yasuo-ozu:
Result: 
clip_test

PR #282, opened 2021-07-04
Add many color variables to color.satyh from Web Color Names

yasuo-ozu:
In latest SATySFi, there are few color variables in color.satyh, and specifying color by RGB is not convenient in some situations.
This change adds color names from Web Color Names.

### DIFF
--- a/lib-satysfi/dist/packages/color.satyh
+++ b/lib-satysfi/dist/packages/color.satyh
@@ -9,15 +9,300 @@ module Color : sig
   val orange : color
   val blue   : color
 
+  % web color names
+  val aliceblue           : color
+  val antiquewhite        : color
+  val aqua                : color
+  val aquamarine          : color
+  val azure               : color
+  val beige               : color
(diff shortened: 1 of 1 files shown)


TonalidadeHidrica:
FYI, [`satysfi-base`](https://github.com/nyuichi/satysfi-base/blob/master/src/color-ext.satyg) has a module that provides web colors.  I also once wanted these wide variety of colors, but now I am wondering if it should be provided by a standard library or an external one.

yasuo-ozu:
@TonalidadeHidrica I am for your standing, so another options is adding famous colors only, such that 
- brown
- cyan
- green
- magenta
- pink
- purple

TonalidadeHidrica:
Agree.  At very least I think "green" should be in the list, since it is one of the most basic color of the additive color model, the absence of which quite astonished me.

yasuo-ozu:
Thanks for discussion, I will edit this PR.

PR #285, opened 2021-07-16
attach doc-comments for primitives

leque:
This patch attaches doc comments to OCaml counterparts of SATySFi primitives. We can generate the documentation by running `dune build \@doc-private`. Generated documents are shown below.

image


gfngfn:
Confirmed that the modified implementation works. Thank you!

Issue #287, opened 2021-07-29
`depext` dependency looks obsoleted


yasuo-ozu:
`depext` package is frozen and it cannot be installed with latest opam (opam=2.1.0).

The features of depext are integrated in opam itself.

https://github.com/ocaml/opam/issues/3790

yuhr:
Due to this incompatibility it seems we cannot install SATySFi for now.

```
❯ opam depext satysfi satysfi-dist satyrographos
opam: unknown command `depext'.
Usage: opam COMMAND ...
Try `opam --help' for more information.

❯ opam install satysfi satysfi-dist satyrographos


<><> Synchronising pinned packages ><><><><><><><><><><><><><><><><><><><><>  🐫 
[satyrographos.0.0.2.10] synchronised (no changes)

[NOTE] Package satyrographos is already installed (current version is 0.0.2.10).
[ERROR] Package conflict!
  * Missing dependency:
    - depext
    unmet availability conditions, e.g. 'opam-version >= "2.0.0~beta5" &
      opam-version < "2.1"'

No solution found, exiting
```

PR #289, opened 2021-09-05
Fix CI with removing unneeded dependency on depext and updating setup-ocaml action

na4zagin3:
SATySFi declares dependency on depext while it in fact does not depend on that; it should be removed.
Furthermore, this unneeded dependency declaration caused issue https://github.com/gfngfn/SATySFi/issues/287

This patch has been backported to past versions: https://github.com/na4zagin3/satyrographos-repo/pull/370

Additionally, let's use `ocaml/setup-ocaml@v2`.

Close https://github.com/gfngfn/SATySFi/issues/287

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,22 +13,40 @@ jobs:
           - 4.10.1
           - 4.11.1
     runs-on: ${{ matrix.os }}
+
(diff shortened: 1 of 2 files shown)


gfngfn:
Thank you so much!

PR #290, opened 2021-09-23
Add char

puripuri2100:
I added `char` type and some functions.

The `char` type uses `Uchar.t` type for implementation.

ref: https://github.com/gfngfn/SATySFi/projects/1#card-54377937

List of added functions:
- `char-to-string : char -> string`
- `char-to-unicode-point : char -> int`
- `char-of-unicode-point : int -> char`
- `char-same : char -> char -> bool`

See the `tests/char.saty` file for how to use them.


puripuri2100:
🤔🤔🤔

```
### output ###
# Error: Unable to set up the cache root directory:
# Unix.Unix_error(Unix.EEXIST, "mkdir",
# "/Users/runner/.cache/dune/db/files/v4")
```
https://github.com/gfngfn/SATySFi/runs/3682988795?check_suite_focus=true#step:5:5926

na4zagin3:
`*-unicode-point` should be renamed for `*-unicode-scalar-value` because “Unicode point” is not a term (See “[Unicode scalar value](https://unicode.org/glossary/#unicode_scalar_value)”).  Did you mean “Unicode [code point](https://unicode.org/glossary/#code_point)”)?.

na4zagin3:
This is my just two cents.

Can this be implemented as a macro like `` ~(char @`2`) `` with `char : input-position * string -> char`?

Otherwise, is it possible to introduce more generic literal syntax (like Scala’s [String Interpolation](https://docs.scala-lang.org/overviews/core/string-interpolation.html), C++’s [User defined literals](https://en.cppreference.com/w/cpp/language/user_literal), Lisp-families’ [read macros](https://gist.github.com/chaitanyagupta/9324402)) instead of one specific to `char`?

For example, define a new syntax `@⟨ident⟩⟨string-literal⟩` (e.g., `` @char`a` ``) which will be parsed as `~(⟨ident⟩ @⟨string-literal⟩)` where `⟨ident⟩` should be a function with type `input-position * string -> char`.  If the `string` is not valid, the function will call `abort-with-message`.

na4zagin3:
Another option would be “not to introduce a literal syntax at all”.  I believe casual users shouldn’t care what is Unicode Scalar Values.  I'm not a fan of user-facing APIs requiring char arguments; they will likely not consider Unicode equivalence.

puripuri2100:
Thank you for pointing out the naming of the primitive functions.

I explain why I introduced the literal syntax of the `char` type.
I would like to use the `char` type when parsing strings. So I expect the `char` type to have two properties:
- Guaranteed to be exactly one character
- Pattern matching is possible

Currently, I must do pattern matching with Unicode scalar value:
https://github.com/puripuri2100/SATySFi-json/blob/master/src/json.satyg#L78

na4zagin3:
Hmm, then can we introduce a macro function `char : string -> int` and make it available at matching clauses?

```satysfi
val f x =
  match x with
  | ~(char `/`) -> lex-string (str-stack^`"`) line (column + 2) ys
```

or with a new macro syntax `@⟨ident_tag⟩⟨string-literal⟩`,

```satysfi
val f x =
  match x with
  | @char`/` -> lex-string (str-stack^`"`) line (column + 2) ys
```

Otherwise, we can extend matching with [view patterns](https://gitlab.haskell.org/ghc/ghc/-/wikis/view-patterns)

```satysfi
val char c =
  match string-length c with
  | 0 -> ``
  | 1 -> string-sub c 0 1
  | _ -> ``
  end

val f x =
  match x with
  | (char -> `\`) -> lex-string (str-stack^`"`) line (column + 2) ys
```

or [extractors](https://docs.scala-lang.org/tour/extractor-objects.html)

```satysfi
val match Char c =
  match string-length c with
  | 0 -> None
  | 1 -> Some (string-sub c 0 1)
  | _ -> None
  end

val f x =
  match x with
  | Char(`\`) -> lex-string (str-stack^`"`) line (column + 2) ys
```

PR #303, opened 2021-11-02
pprint Range.Normal in compact (GNU style) format

leque:
This PR changes pprint format of `Range.Normal` to GNU style `file:line.pos1-pos2` format[^1] to reduce characters and for ease of tool support, e.g. jumping to a location. This affects test results only.

[^1]: https://www.gnu.org/prep/standards/html_node/Errors.html#Errors

### DIFF
--- a/src/frontend/range.ml
+++ b/src/frontend/range.ml
@@ -1,7 +1,19 @@
 
+(*
+   format ranges in GNU style format.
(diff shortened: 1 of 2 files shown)


gfngfn:
I’m so sorry for the late response. Thanks!

PR #304, opened 2021-11-03
Fixed the value of tDASH

puripuri2100:
It was reported that an error occurred.
https://github.com/amutake/docker-satysfi/pull/65#issuecomment-957089698

The cause is that the definition of `tDASH` has changed.

I fixed this.


gfngfn:
Thank you for fixing! (I understand the origin of the bug at almost the same time!)

leque:
Thanks for the fix. BTW are there any chance to catch this sort of errors in CI?

gfngfn:
Probably #197 (which has been left unmerged before discarding the Travis CI workflow) can be used as a reference for developing CI.

gfngfn:
Oops, editing the CI workflow before merging this PR would have been better.

gfngfn:
To be sure, I confirmed that after installing the modified implementation running `cd doc && make` produces `primitives.pdf` correctly.

PR #305, opened 2021-11-06
run tests on GitHub Actions

leque:
This PR is a reboot of #96 and #197. This runs `dune runtest` and then build some documents by using installed SATySFi on the CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,3 +51,18 @@ jobs:
 
       - name: Build SATySFi
         run: opam exec -- make all
+
+      - name: Run tests
+        run: |
+          opam exec -- dune runtest
+
+      - name: Install SATySFi
+        run: |
+          opam install -y .
+          ./download-fonts.sh
+          sudo ./install-libs.sh
+
+      - name: Build demo docs
+        run: |
+          (cd demo; opam exec -- make)
+          (cd doc; opam exec -- make)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,14 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
 
+      - name: Cache dist fonts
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -51,3 +51,18 @@ jobs:
...
         run: opam exec -- make all
+
+      - name: Run tests
+        run: |
+          opam exec -- dune runtest
+
+      - name: Install SATySFi
+        run: |
+          opam install -y .
+          ./download-fonts.sh
```
na4zagin3, line 62: Can you cache the downloaded fonts (see https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows)?  The other option is to have OPAM cache downloaded files (#149).

When I was using Travis CI, running download-fonts.sh without caching had the CI instances banned from some font distributing sites.

.github/workflows/ci.yml
```diff
@@ -51,3 +51,18 @@ jobs:
...
+          sudo ./install-libs.sh
```
na4zagin3, line 63: This is a matter of taste.  You can install the lib files under `~/.satysfi` instead.

```
./install-libs.sh ~/.satysfi
```

leque: fair enough!

.github/workflows/ci.yml
leque: Thank you. I have added cache settings.

leque:
Updated. Now we install fonts into `~/.satysfi`, and use caches to avoid downloading fonts on each build.

.github/workflows/ci.yml
na4zagin3: Thank you!

gfngfn:
Sorry for the late response… I understand that caching of fonts is enabled and that it seems indeed working. Thank you for the improvement!

PR #316, opened 2022-01-12
Fix description of register-document-information

puripuri2100:


gfngfn:
Good catch! By consulting `primitives.cppo.ml` I understand that the modified implementation seems correct:

```
let tDOCINFODIC =
  let asc =
    Assoc.of_list [
      ("title", tOPT tS);
      ("subject", tOPT tS);
      ("author", tOPT tS);
      ("keywords", tL tS);
    ]
  in
    (~! "document-information-dictionary", RecordType(asc))
```

Thank you for fixing.

PR #317, opened 2022-01-16
Suggestion to remove memo-ja-how-to-use.md

AumyF:
[/memo-ja-how-to-use.md](https://github.com/gfngfn/SATySFi/blob/master/memo-ja-how-to-use.md) is linked as “使用方法” on [Wiki](https://github.com/gfngfn/SATySFi/wiki/SATySFi-Wiki) but the file does not explain "how to use" but "how library roots works".  Furthermore, it has not been updated since 2018.

In my opinion, it should be rewritten as a page of the Wiki to make it easy to update.


gfngfn:
That sounds reasonable. Thank you for suggestion.

PR #320, opened 2022-03-22
fix opam

yasuo-ozu:
Fix build of SATySFi.

https://twitter.com/yasuo_ozu/status/1506258738323521538


gfngfn:
Sorry to reply late. Thank you for fixing!

PR #330, opened 2022-07-30
friendlier error message for empty files

### LINKED ISSUE #270: Inputting an empty file leads to an unfriendly message
If we input `satysfi` an empty file, it emits the following error message:

```
 ---- ---- ---- ----
  target file: 'report1.pdf'
  dump file: 'report1.satysfi-aux' (will be created)
  parsing 'report1.saty' ...
! [Syntax Error at Parser] at "report1.saty", line 1, characters 0-0:
Compile error.
```

It may be nicer if it outputs differently for specifically an empty file.

leque:
This PR attempts to fix #270.

Error message will be changed as below (last line):
```console
$ touch e.saty
$ ./satysfi e.saty
 ---- ---- ---- ----
  target file: 'e.pdf'
  dump file: 'e.satysfi-aux' (will be created)
  parsing 'e.saty' ...
! [Syntax Error at Parser] at "e.saty", line 1, characters 0-0:
    empty input
```


gfngfn:
Confirmed that the modified version works. Thanks!

Issue #332, opened 2022-08-05
Awkward spacing at the beggining of a math equation


sankantsu:
When you write a binary operator (such like `+` or `-`) at the beggining of a math equation, the spacing between binary operator and the following number looks awkward.

For example, this code shows this awkward spacing.

```
@require: stdjareport

document (|
    title = {Awkward spacing at the begging of math};
    author = {sankantsu};
|) '<
     +p {
         ${+1 - 2 = -1}
     }
>
```

As a user-level solution, you can avoid this spacing problem by defining a math-command to treat `+` or `-` as MathOrd class using `math-char` primitive and use this command at the beggining of the equation.

I think when a `+` or `-` symbol is put at the beggining of a math equation it is usually intended to be an unary operator for the following number.

I don't know whether this behavior of the SATySFi is intentional or not, but TeX seems to be treating binary operators at the beggining of a equation as a Ord symbol and I prefer TeX's approach.

It seems that very small change to the source code fix this problem.
Please consider changing this spacing behavior.

gfngfn:
Merged #333. Thank you!

PR #333, opened 2022-08-05
Fix spacing of binary operator at the beggining of math equation (issue #332)

sankantsu:
This change fix spacing of a binary operator at the beggining of math equation, treating a binary operator at the beggining of a equation as a MathOrd symbol.

The change involves only one line addition in `src/frontend/math.ml`.

More detailed explanation is in issue #332.

### DIFF
--- a/src/frontend/math.ml
+++ b/src/frontend/math.ml
@@ -257,6 +257,7 @@ let normalize_math_kind mkprev mknext mkraw =
         | MathRelation
         | MathOpen
         | MathPunct
+        | MathEnd
             -> true
         | _ -> false
       in


gfngfn:
Thank you for sending a PR!

> I don't know whether this behavior of the SATySFi is intentional or not, but TeX seems to be treating binary operators at the beggining of a equation as a Ord symbol and I prefer TeX's approach.
>
> (#332)

This is not intentional, although I did notice that the strange behavior of spacing happens sometimes.

The modification looks completely right, and I confirmed that it indeed works:

```
@require: stdjareport

let-math \unary-plus = math-char MathOrd `+`

in
document (|
    title = {Awkward spacing at the beginning of math};
    author = {sankantsu};
|) '<
     +p {
         ${+1 - 2 = -1}
     }
     +p {
         ${\unary-plus 1 - 2 = -1}
     }
>
```

- Before:

  

- After:

  

Thank you so much for looking into and fix the implementation!

PR #339, opened 2022-08-14
Add `\lnot` to math.satyh

SnO2WMaN:
It seems counter-intuitive to have `\land`, `\lor` and no `\lnot`.

### DIFF
--- a/lib-satysfi/dist/packages/math.satyh
+++ b/lib-satysfi/dist/packages/math.satyh
@@ -812,7 +812,7 @@ end = struct
   let-math \bigcap = bigop `⋂`
   let-math \bigcup = bigop `⋃`
 
-  let-math \lnot = rel `¬`
+  let-math \lnot = ord `¬`
   let-math \land = rel `∧`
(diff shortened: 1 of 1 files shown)


lib-satysfi/dist/packages/math.satyh
```diff
@@ -811,6 +812,7 @@ end = struct
   let-math \bigcap = bigop `⋂`
   let-math \bigcup = bigop `⋃`
 
+  let-math \lnot = rel `¬`
```
gfngfn, line 815: Thanks for improving! Since ¬ is a unary operator, `ord` would be suitable here instead of `rel`.

gfngfn (changes requested):

lib-satysfi/dist/packages/math.satyh
SnO2WMaN: fix this. thanks

gfngfn (approved):

gfngfn:
Merged. Thank you!

PR #364, opened 2022-10-09
Refactor the implementation of the typechecker into monadic one

gfngfn:


src/myUtil.ml
```diff
@@ -104,6 +104,26 @@ module ResultMonad = struct
   let return v = Ok(v)
 
   let err e = Error(e)
+
+  let ( let+ ) = ( >>= )
```
leque, line 108: `(let*)` for `(>>=)` would be conventional.

https://v2.ocaml.org/manual/bindingops.html#ss%3Aletops-rationale

gfngfn: Thanks! I will replace `let+` with that.

PR #365, opened 2022-10-13
add syntax error with trailing optional arguments

anko9801:
For example, in this case, there is no way to determine whether a type of `f {a}` is `?B -> Ret` or `Ret`. 
```
let f a ?:b = b in
! [Type Error] at ...:
    undefined variable 'b'.
```
So far, The trailing optional arguments have been ignored. This PR fixed to show errors when parsing.
```
let f a ?:b = b in
! [Syntax Error at Parser] at ...:
    non-optional argument is needed at the end.
```


gfngfn:
Looks good. Thank you for fixing!

PR #367, opened 2022-10-19
Refactor `typechecker.ml` by dividing it

gfngfn:


src/frontend/signatureSubtyping.mli
```diff
@@ -2,8 +2,17 @@
 open StaticEnv
 open TypecheckUtil
 
+(** [substitute_concrete θ ξ] means [θ(ξ)],
```
leque, line 5: `substitute_abstract`?

gfngfn: Thanks!

PR #388, opened 2022-12-18
Support fixed dependencies

gfngfn:


gfngfn:
Achieved by https://github.com/gfngfn/SATySFi/pull/429 instead.

Issue #359, opened 2022-09-22
record punning


leque:
I would like to be able to use so-called record punning found in [Successor ML](https://github.com/SMLFamily/Successor-ML/wiki/Summary-of-proposed-changes#records), [OCaml](https://v2.ocaml.org/manual/expr.html#sss:expr-records), [ECMAScript (Shorthand property names)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer), ..., etc. also in SATySFi.

With this feature, we can construct record
```
let f x y = (| x; y |)
```
instead of
```
let f x y = (| x = x; y = y |)
```

This can be also applied to record patterns (#358), so we can write
```
let f (| x; y |) = (x, y)
```
instead of
```
let f (| x = x; y = y |) = (x, y)
```

PR #389, opened 2023-01-03
record punning & record patterns

### LINKED ISSUE #358: record patterns
It would be useful if SATySFi supports record patterns.

In SML#, the first pattern in the example below matches a record having the specified set of labels, and the second pattern matches any record containing the set of specified labels. ([8.3 Record patterns‣ Chapter 8 SML# feature: record polymorphism ‣ Part II Tutorials ‣ SML# Document Version 4.0.0 - SML# Project](https://smlsharp.github.io/en/documents/4.0.0/Ch8.S3.html))

```sml
# fun f {X = x, Y = y} = (x, y);
val f = fn : [’a, ’b. {X: ’a, Y: ’b} -> ’a * ’b]
# fun f {X = x, Y = y, ...} = (x, y);
val f = fn : [’a#{X: ’b, Y: ’c}, ’b, ’c. ’a -> ’b * ’c]
```

I think SATySFi can follow the same strategy: e.g. (in 0.0.x syntax)

```
% val f: (| x : 'a; y : 'b |) -> 'a * 'b
let f (| x = x; y = y |) = (x, y)

% val g: 'r -> 'a * 'b constraint 'r :: (| x : 'a; y : 'b |)
let g (| x = x; y = y; _ |) = (x, y)
```

leque:
resolves #358 and #359.

This PR implements record punning and record patterns in 0.1.0 branch.

Record punning is a short-hand notation for record construction. You can omit an field value if it is a variable reference and it has the same name as a record field. E.g. `(| x = x |)` can be written as `(| x |)`.

Record pattern enables us to pattern match against record values. Record patterns are typed as described in #358:

```
  % fx : (| x : 'a |) -> 'a
  val fx (| x |) = x

  % fx- : (| x : 'a | ?'b |) -> 'a
  val fx- (| x, _ |) = x
```

Some remarks:

- `--bytecomp` is not tested since 0.1.0 does not support `eval_libary_file` in bytecomp mode yet.
- Exhaustiveness check is not implemented because I cannot understand what algorithm is used in `Exhckecker`. It would be nice to be implemented as a separated PR.


src/frontend/parser.mly
```diff
@@ -1055,6 +1061,31 @@ pattern_non_var_bot:
...
             make_standard (Tok tokL) (Tok tokR) (UTPTuple(TupleList.make utpat1 utpat2 utpat_rest))
       }
+  | tokL=L_RECORD; body=pat_record_body; tokR=R_RECORD
+      {
+        let (is_open, fields) = body in
+        make_standard (Tok tokL) (Tok tokR) (UTPRecord(is_open, fields))
+      }
+;
+pat_record_body:
+  | WILDCARD; COMMA?
```
gfngfn, line 1071: The description `COMMA?` allows `(| foo = x, _, |)` for a record pattern as well as `(| foo = x, _ |)`. Is this intended?

leque: Yes, this is intentional. I think there is no reason to forbid a trailing comma after a wildcard.

Et7f3: Same for list `[ value1; ]` python also accept trailing comma: one advantage is we can swap item and it will just works. Other reason seems to be human better deal with terminator than separator (last argument was given by an ocp-indent author).

PR #391, opened 2023-01-09
fix error of `\nLeftarrow` and `\nRightarrow`

takemar:
their definitions were swapped.


gfngfn:
Good catch! Thank you.

PR #417, opened 2023-10-03
Add split-grapheme-cluster and string normalization function

puripuri2100:
- Add `split-grapheme-cluster : string -> string list`
- String normalization when parsing strings and inline-text

### DIFF
--- a/satysfi.opam
+++ b/satysfi.opam
@@ -34,6 +34,7 @@ depends: [
   "ppx_deriving"
   "re" {build}
   "uutf"
+  "uunf"
   "yojson-with-position" {= "1.4.2+satysfi"}
   "omd" {< "2.0.0~"}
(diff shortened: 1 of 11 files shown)


gfngfn:
Thanks for making a PR!

I think the following API for strings may be a better formalization:

- `val normalize-string-to-nfc : string -> string`
- `val normalize-string-to-nfd : string -> string`

in that this:

- allows users to use other normalization forms than NFC, such as NFD, and
- does not require that string values are always internally NFC, and thereby keeps the implementation less error-prone

Would you agree with this kind of API design?

puripuri2100:
I agree it. I fixed the code.

gfngfn:
I've re-read this PR and found it to be basically good. Thank you!

PR #419, opened 2023-10-22
Generate `version.ml` from `satysfi.opam` to make releasing easy

y-yu:
- When releasing (for example #418), the releaser(= gfn-san) rewrite the version of `src/frontend/main.ml` by their hands
- `tools/genversion` will generate such a `version.ml` 👇  when building
    ```ocaml
    let version = "0.0.9"
    ```
    and `genversion` gets version string (`"0.0.9"`) from `satisfy.opam` using my ad-hoc regex
- Then `src/frontend/main.ml` uses the generated version string
- I think that this PR reduces a step of releasing process 🤔


na4zagin3:
You can have OPAM run substitute commands like https://github.com/na4zagin3/satyrographos/pull/189 (`dune subst` and ‘sed`)

y-yu:
Thanks for the comment 😄 I saw the technique of https://github.com/na4zagin3/satyrographos/pull/189.
First of all, the best choice is what gfn-san likes, I think. Based on that, my opnions about this PR (using OCaml way) _v.s._ `dune` and `sed`(zagin-san's way) are: 

- zagin-san's way is simpler than my PR way
- In `sed`, there are some differences between BSD and GNU so I don't like to use `sed`😇 
    - Indeed, on this usecase, `sed` does simple replacement only so is it no matter? 🤔 
    - OCaml is more complecate than `sed` but it's the same behavior on BSD and GNU
- OS commands and shell scripts are not unit testable

y-yu:
Even if we'll choice either way, it would be better to write the version info to the other file (such like `version.opam`) rather than using single OPAM file 🤔 in order to manipulate it by GitHub Actions(automatic release scripts) or so on. (but it's not possible on the current OPAM)
It may be a lesson for SATySFi 0.1.0 package manager setting file 👀

gfngfn:
I'm so sorry for the late response.

I can't confidently tell which of the two approaches is better, but at least I like Yoshimura-san's OS-independent implementation. I would thus like to merge this PR for the moment.

na4zagin3:
I'm so sorry for my long silence.  Avoiding using `sed` achieves OS-independency obviously, which is nice!  This is too late but anyway I retract my comment.  Thank you!

PR #421, opened 2023-12-02
Allow importing submodules in use statement

kyori19:
Currently, in order to open submodules we need to combine `use Module` and `let open SubModule in`.
To keep consistency and make headers more simple, allow writing `use Module.SubModule` to import submodules directly.

### DIFF
--- a/demo/local.satyh
+++ b/demo/local.satyh
@@ -1,15 +1,12 @@
-use package Stdlib
+use package open Stdlib.List
+use package open Stdlib.Color
+use package open Stdlib.HDecoSet
+use package open Stdlib.VDecoSet
 use package Math
 use package Code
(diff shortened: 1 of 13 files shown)


demo/local.satyh
```diff
@@ -1,16 +1,13 @@
-use package Stdlib
+use package open Stdlib.List
```
gfngfn, line 1: The description `use package open Stdlib.List` looks like adding all identifiers in `Stdlib.List` (such as `map`, `fold-left`, etc.) to the current environment. I guess your intention is different; only `List` itself is wanted to be added to the current environment. It would thus be better to write `use package Stdlib.List` (i.e., to remove `open`).

gfngfn:
Thank you very much!

PR #435, opened 2024-03-25
Separate `eval_document_file` to `BuildDocument` module

kyori19:
* Removed dependency to `CrossRef`, `PageBreak` from `main.ml`
* Organized some codes which handles `CrossRef`
* Extracted differences between output modes to 3 steps: `transform`, `reset` and `output`


gfngfn:
Your changes look great to me. I ran the refactored version locally and confirmed that it indeed works fine for section numbering, etc. Thanks!

PR #438, opened 2024-04-10
Update GitHub Actions

smorimoto:
https://github.com/smorimoto/SATySFi/actions/runs/8638386054


gfngfn:
Sorry for the late review :bow: I Looked into the modified description and found that it's good (according to the documentation of `cache@v4`). Thank you very much!

PR #442, opened 2024-06-21
Support transparent PNG images

nsfisis:
# Overview

It will add support for transparent PNG images (images that have alpha channel).

Here is the screenshot of `tests/images/test.pdf` generated from `tests/images/test.saty`:

image

The last 4 images have alpha channel and newly supported by the PR.

# Implementation details

Extract the raw byte sequence of the alpha channel and create an XObject for the alpha mask. Then, add `/SMask` field that refers to the created XObject to the main image's dictionary.

### DIFF
--- a/src/backend/loadRawImage.ml
+++ b/src/backend/loadRawImage.ml
@@ -3,37 +3,14 @@ open MyUtil
 type data =
   { raw_bytes : bytes
   ; bits_per_components : int
+  ; alpha_bytes : bytes option
   }
 
 let cannot_load_image msg abs_path = raise @@ ImageHashTable.CannotLoadImage (msg, abs_path)
 
-let make_xobject (pdfmain : Pdf.t) (colorspace : Pdf.pdfobject) widdots hgtdots abs_path =
-  let img =
-    try Images.load (get_abs_path_string abs_path) []
-    with
-    | Failure message -> cannot_load_image message abs_path
-  in
-  let {raw_bytes; bits_per_components} =
-    match img with
-    | Rgb24 img ->
-      { raw_bytes = Rgb24.dump img
-      ; bits_per_components = 8
-      }
-    | Index8 img ->
-      { raw_bytes = Rgb24.dump (Index8.to_rgb24 img)
-      ; bits_per_components = 8
-      }
-    | Index16 img ->
-      { raw_bytes = Rgb24.dump (Index16.to_rgb24 img)
-      ; bits_per_components = 8
-      }
-    | Rgba32 _ ->
-      cannot_load_image "images with alpha channel are not supported" abs_path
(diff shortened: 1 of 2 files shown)


gfngfn:
Thank you very much for the great improvement!

Confirmed that:

- the modified implementation works as intended, and
- the implementation conforms to the specification of PDF 1.7 by consulting the spec:

> ### 8.9.5 Image Dictionaries
>
> …
>
> | Key     | Type   | Value |
> |---------|--------|-------|
> | … | … | … |
> | `SMask` | stream | (Optional; PDF 1.4) A subsidiary image XObject defining a soft-mask image (see 11.6.5.3, "Soft-Mask Images") that shall be used as a source of mask shape or mask opacity values in the transparent imaging model. The alpha source parameter in the graphics state determines whether the mask values shall be interpreted as shape or opacity. If present, this entry shall override the current soft mask in the graphics state, as well as the image’s `Mask` entry, if any. However, the other transparency-related graphics state parameters—blend mode and alpha constant—shall remain in effect. If `SMask` is absent, the image shall have no associated soft mask (although the current soft mask in the graphics state may still apply).
> | … | … | … |

> #### 11.6.4.3 Mask Shape and Opacity
>
> …
>
> - An image XObject may contain its own soft-mask image in the form of a subsidiary image XObject in the `SMask` entry of the image dictionary (see “Image Dictionaries”). This mask, if present, shall override any explicit or colour key mask specified by the image dictionary’s `Mask` entry. Either form of mask in the image dictionary shall override the current soft mask in the graphics state.

> #### 11.6.5.3 Soft-Mask Images
>
> The second way to define a soft mask is by associating a soft-mask image with an image XObject. This is a subsidiary image XObject specified in the `SMask` entry of the parent XObject’s image dictionary (see “Image Dictionaries”). Entries in the subsidiary image dictionary for such a soft-mask image shall have the same format and meaning as in that of an ordinary image XObject (as described in Table 89 in “Image Dictionaries”), subject to the restrictions listed in Table 145.

> Table 145 – Restrictions on the entries in a soft-mask image dictionary
>
> | Key | Restriction |
> |-----|-------------|
> | `Type` | If present, shall be `XObject`. |
> | `Subtype` | Shall be `Image`. |
> | `Width` | If a `Matte` entry (see Table 146) is present, shall be the same as the `Width` value of the parent image; otherwise independent of it. Both images shall be mapped to the unit square in user space (as are all images), regardless of whether the samples coincide individually. |
> | `Height` | Same considerations as for `Width`. |
> | `ColorSpace` | Required; shall be `DeviceGray`. |
> | `BitsPerComponent` | Required. |
> | `Intent` | Ignored. |
> | `ImageMask` | Shall be `false` or absent. |
> | … | … |

PR #445, opened 2024-07-03
Prepare for Better Windows Support

smorimoto:
This PR updates setup-ocaml to v3. Some disabled Windows pieces allow you to run tests on Windows just by enabling them when dependencies support Windows in the _near_ future.


.github/workflows/ci.yml
```diff
@@ -7,42 +7,42 @@ on:
...
     strategy:
       fail-fast: false
       matrix:
         os:
           - ubuntu-latest
           - macos-latest
-        ocaml-compiler:
-          - "4.12"
-          - "4.13"
-          - "4.14"
```
smorimoto, line 19: OCaml 4.13 or later is supported on Windows. It could be improved in the future, but honestly, testing on 4.14 is good enough.

gfngfn: Thank you!

How about keeping this array as the following singleton?

```yaml
ocaml-compiler:
- "4.14"
```

I think this will ease the future addition of other versions.

smorimoto: Sure!

.github/workflows/ci.yml
```diff
@@ -56,11 +62,29 @@ jobs:
...
+  lint-opam:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout tree
+        uses: actions/checkout@v4
+
+      - name: Set-up OCaml
+        uses: ocaml/setup-ocaml@v3
+        with:
+          ocaml-compiler: "4.14"
```
gfngfn, line 83: Should this be `"4.14"` instead of `${{ matrix.ocaml-compiler }}` for the purpose of linting? I'm not sure which is more appropriate 😅

smorimoto: You don't need to have multiple compiler versions except for build and test, so you just need to specify the latest supported compiler version!

gfngfn: Ok, thanks!

gfngfn:
Thank you very much!

PR #446, opened 2024-07-04
`Pervasives` -> `Stdlib`

smorimoto:
The `Pervasives` module is no longer included in OCaml 5.0, and `Stdlib` is the recommended replacement. `Stdlib` itself has been included since OCaml 4.07 and is stable enough.

### DIFF
--- a/src/backend/fontFormat.ml
+++ b/src/backend/fontFormat.ml
@@ -369,14 +369,14 @@ let per_mille ~(units_per_em : int) (w : design_units) : per_mille =
 module GSet = Set.Make
   (struct
     type t = original_glyph_id
-    let compare = Pervasives.compare
+    let compare = Stdlib.compare
   end)
 
 
 module GMap = Map.Make
   (struct
     type t = original_glyph_id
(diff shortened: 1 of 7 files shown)


smorimoto:
This should allow you to add support for OCaml 5.2 by just updating your forked `camlpdf`!

smorimoto:
The build error occurs because the certificate for gust.org.pl has expired, but is not directly related to this PR 🤷‍♂️

smorimoto:
It might be a good idea to create a repository like this [ocaml/opam-source-archives](https://github.com/ocaml/opam-source-archives) and archive font files etc. in a stable place.

You can still download the font file here: https://web.archive.org/web/https://www.gust.org.pl/projects/e-foundry/latin-modern/download/lm2.004otf.zip

smorimoto:
@gfngfn Upstream issue is resolved!

smorimoto:
Gentle ping

gfngfn:
Thanks! (& sorry for the late response :pray:)

PR #447, opened 2024-07-04
Cleanup

smorimoto:
- Remove preventative upper bounds for dependencies
  - https://github.com/ocaml/opam-repository/wiki/Policies#5-strict-dependency-constraints-should-be-avoided---or-preventive-upper-bounds
- Replace `genversion` with `dune-build-info`
- Replace the Makefile-based one with the dune-based one
  - Delegate opam installation to dune (See `satysfi.opam`)


smorimoto:
Ready to merge!

smorimoto:
Instead of genversion, you can use dune-build-info to drop a lot of code 🙃

smorimoto:
Gentle ping

.github/workflows/ci.yml
```diff
@@ -42,7 +42,6 @@ jobs:
         uses: ocaml/setup-ocaml@v3
         with:
           ocaml-compiler: ${{ matrix.ocaml-compiler }}
-          dune-cache: true
```
gfngfn, line 45: Sorry for the very late response! :bow: (I have been tied up with other tasks for a while.)

Let me check for sure: is the removal of `dune-cache: true` a workaround for Windows?

dune-project
```diff
@@ -28,51 +26,33 @@
     (>= 4.12.0)
     (< 5.0.0)))
   (batteries
-   (and
-    (>= 3.6.0)
-    (< 4.0)))
+   (>= 3.6.0))
```
gfngfn, line 29: @smorimoto Thank you for citing the following page, I've read it:

> * Remove preventative upper bounds for dependencies
>   - https://github.com/ocaml/opam-repository/wiki/Policies#5-strict-dependency-constraints-should-be-avoided---or-preventive-upper-bounds

I think I understand the description of the policies on that page itself. Still, I couldn't find what will indeed happen when a _backward incompatible_ version of some dependency is released. Specifically, I need clarification on whether OPAM or I is responsible for resolving the incompatibility at that moment. (If it's not OPAM but me,  removing "preventive upper bounds" does not sound very reasonable to me.) Is that documented somewhere else?

.github/workflows/ci.yml
smorimoto: Oops

dune-project
smorimoto: @gfngfn Basically, the responsibility lies with the maintainer of the opam repository as long as it's published in the opam repository. If not, it's generally considered the best way to cron GitHub Actions and regularly check that it's not broken. (So basically, repository owner, you're responsible.)
If this doesn't feel very reasonable, it probably makes sense to put upper bounds, but it's widely known that bulk dependency updates are more painful than gradual updates, so it's kind of a trade-off.

.github/workflows/ci.yml
smorimoto: Fixed.

smorimoto: It should be available on all operating systems btw!

dune-project
gfngfn: Thank you for answering :pray: but I'm not very sure what the following means, unfortunately:

> it's widely known that bulk dependency updates are more painful than gradual updates

It's true that constraint solving can be more feasible without upper bounds than with upper bounds, but I don't think this implies that we shouldn't use upper bounds for backward compatibility; it will just incur the possibility of installing dependencies with interfaces that are unintended for users.

My original concern was that there will always be a possibility that a new backward-incompatible release of some dependencies breaks _already released versions_ of SATySFi that have no upper bounds. Indeed, breakdowns of this kind have happened a few times so far. (As far as I remember, backward-incompatible releases of `omd`, `camlimages`, and `core_kernel` have caused ones.)

Owing to this concern, I think it would be better to keep upper bounds, at least for our project. Does this make sense to you?

smorimoto: We can consider it reasonable. Anyway, as the owner, you always have the right to decide everything!

smorimoto: Reverted upper bound changes!

gfngfn: Thank you very much for the update!

Btw, I'm a bit curious about what brought the OCaml community to the idea of avoiding upper bounds. It might be better for me to follow the original discussion in the community at some point.

.github/workflows/ci.yml
```diff
@@ -71,20 +71,30 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
       - name: Checkout tree
         uses: actions/checkout@v4
-
       - name: Set-up OCaml
         uses: ocaml/setup-ocaml@v3
         with:
-          ocaml-compiler: "4.14"
-          dune-cache: true
```
smorimoto, line 84: Since a dune cache will be created for each job, you may want to set it only for the main job to avoid cache thrashing due to GitHub's cache size limit.

smorimoto:
@gfngfn Ready to merge!

.github/workflows/main.yml
```diff
@@ -19,7 +16,7 @@ jobs:
           - macos-latest
           # - windows-latest
         ocaml-compiler:
-          - "4.14"
+          - 4
```
smorimoto, line 19: Specify 4 instead of 4.14 to let setup-ocaml always lookup the latest OCaml 4 variant!

gfngfn:
Thank you so much for arranging!

gfngfn:
Oops, merging failed and the CI workflows started to run again :confused:

src/version.ml
```diff
@@ -0,0 +1,4 @@
+let version =
+  match Build_info.V1.version () with
+  | None -> "dev"
+  | Some v -> Build_info.V1.Version.to_string v
```
smorimoto, line 4: Do not forget to add a version field to the opam file during release process!

smorimoto:
Oops 😅

smorimoto:
Thanks a lot @gfngfn!

gfngfn:
Successfully merged :+1:

PR #448, opened 2024-08-04
Fix Opam Dependency Submission workflow

smorimoto:
Totally my bad...


smorimoto:
@gfngfn

gfngfn:
I didn't noticed that, thank you very much!

smorimoto:
Now we have a dependency graph! https://github.com/gfngfn/SATySFi/network/dependencies

PR #449, opened 2024-08-09
fix duplication of font-latin-bold

ghaaj:
I removed lines 38 and 120 in stdjareport.satyh. They duplicated some of the changes in 25b83af which had been committed by the time #126 was merged.


gfngfn:
You noticed that well. Thank you!

PR #475, opened 2026-03-01
Validate page-content scheme and skip main text on invalid pages

MasWag:
## Summary
- Validate `pagecontf` output per page (`text-origin` inside paper bounds and `text-height > 0pt`).
- If invalid, skip main text rendering on that page, keep header/footer rendering, and defer body contents to following pages.
- Add single-column page-number limit handling, aligned with multicolumn behavior.
- Move page-size to paper-dimension conversion to `HandlePdf` and reuse it from `PageBreak`.

## Changes
- Added `HandlePdf.get_paper_width_and_height` and exported it.
- Removed duplicate paper conversion logic from `PageBreak`.
- Added page-content validation/warning path in both single-column and multicolumn flows.
- Preserved existing behavior for valid page-content.

## Testing
- `dune build`
- `dune build @runtest`


MasWag:
It seems this patch does not work with class files "rendering" outside the pages, such as SLyDIFi (https://github.com/monaqa/slydifi/blob/e9d0f57c9e27c77888582eaa9ad8b9fd35a12828/src/slydifi.satyh#L96).

